PR #81, opened 2021-10-25
feat: open selected message attachments in browser using keybinding

chromium539:
wip, I still need to implement opening attachments of messages with multiple attachments
default keybinding is `o`

### DIFF
--- a/config.go
+++ b/config.go
@@ -24,6 +24,7 @@ type keybindings struct {
 	ReplySelectedMessage        string `toml:"reply_selected_message"`
 	MentionReplySelectedMessage string `toml:"mention_reply_selected_message"`
 	CopySelectedMessage         string `toml:"copy_selected_message"`
+	OpenAttachmentInSelectedMessage string `toml:"open_attachment_in_selected_message"`
 }
 
 type theme struct {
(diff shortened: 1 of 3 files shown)


config.go
```diff
@@ -24,6 +24,7 @@ type keybindings struct {
 	ReplySelectedMessage        string `toml:"reply_selected_message"`
 	MentionReplySelectedMessage string `toml:"mention_reply_selected_message"`
 	CopySelectedMessage         string `toml:"copy_selected_message"`
+	OpenAttachmentInSelectedMessage string `toml:"open_attachment_in_selected_message"`
```
ghost: Do you mind renaming this field to `OpenSelectedMessageAttachments`?

util.go
```diff
@@ -201,3 +204,18 @@ func isUnread(c *discordgo.Channel) bool {
...
 }
+
+func openInDefaultBrowser(url string) error {
+	// source: https://stackoverflow.com/a/10380889/13173909 
+	var err error
+	switch runtime.GOOS {
+		case "linux":
+			err = exec.Command("xdg-open", url).Start()
+		case "windows", "darwin":
+			err = exec.Command("open", url).Start()
```
ghost: Windows should use `start` command instead. You may use a [third-party package](https://github.com/skratchdot/open-golang) to achieve this if you would like.

PR #102, opened 2021-11-18
Fix MentionReplySelectedMessage keybind in ui.go

canisli:
After commit e8a2f33, checking for MentionReplySelectedMessage keystroke incorrectly checked for ReplySelectedMessage instead.


ghost (approved):
Thanks!

PR #137, opened 2022-03-12
feat: implement keybindings for opening and downloading attachments

### LINKED ISSUE #136: Feature request: Quick file attachment downloading/opening
This might already be a thing but it's unclear to me how to access it, it does not seem to appear in any of the keybinds or in the message action menu.

I think it would be useful if simply pressing enter or some key like `d` while hovering over a message with an attachment to download it to the default download location on the system. 

In addition, I think a keybind like `o` to automatically open the file in the default system handler would be pretty neat.

KodiCraft:
This adds a few new config options:

`attachment_download_location` a string which is, by default, set to the platform's default downloads folder.

`open_attachment` a keybind which, when used while hovering over a message with at least one attachment, will open those attachments in the system's default handler. It uses `start` on Windows, `xdg-open` on Linux and `open` on MacOS

`download_attachment` a keybind which, when used while hovering over a message with at least one attachment, will download those attachments in the location configured in `attachment_download_locaiton`

as well as, of course, all the features relating to those configuration options.

While I could test it on Windows and Linux, I am as of yet unsure if this works on MacOS. Also, this does not implement any user feedback upon downloading or opening a file, as I am currently unfamiliar with the UI code.

btw this is my first ever PR so I'd appreciate any feedback


KodiCraft:
Further testing reveals that opening file is still wonky on Windows, I'm unsure why currently. Downloading works but opening doesn't. I'll look into it tomorrow

KodiCraft:
I identified the issue, it's related to the fact that `start` on Windows seems to inconsistently switch between using magic numbers and file extensions. I am going to push a fix today

KodiCraft:
Okay now it's fixed, opening files directly from discordo works for Linux and Windows. Although, and this is an issue with Windows more than anything, if Windows does not have an association for the file it will silently refuse to open it.

.gitignore
```diff
@@ -0,0 +1 @@
+discordo*
```
ghost: Ideally, you should append `.vscode/launch.json` to this file.

```suggestion
discordo*

# Visual Studio Code
.vscode/
```

KodiCraft: I completely forgot about that, my bad

config/general.go
```diff
@@ -11,14 +16,28 @@ type GeneralConfig struct {
 	Mouse              bool           `toml:"mouse"`
 	Timestamps         bool           `toml:"timestamps"`
 	Identify           IdentifyConfig `toml:"identify"`
+	DownloadLocation   string         `toml:"attachment_download_location"`
```
ghost: Renaming this field as well as `toml` tag to `AttachmentDownloadsDir` and `attachment_downloads_dir` is preferable to be more specific.

config/general.go
```diff
@@ -11,14 +16,28 @@ type GeneralConfig struct {
...
+	dlloc := "~/Downloads/"
+	if runtime.GOOS == "windows" {
+		h, _ := os.UserHomeDir()
+		dlloc = h + "\\Downloads\\"
+	} else if runtime.GOOS == "linux" || runtime.GOOS == "darwin" {
+		h, _ := os.UserHomeDir()
+		dlloc = h + "/Downloads/"
+	} else {
+		dlloc = os.TempDir() // Very lame fallback, I know
+	}
```
ghost: - Extract this into a new function `func UserDownloadsDir() string`
- Remove the trailing slash at the end of the paths since this is later added using `filepath.Join()`; since, `os.TempDir()` returns path without a trailing slash, `os.Create()` call while downloading the file will return `/tmp<attachment.name>` instead of `/tmp/<attachment.name>`.

ui/messages.go
```diff
@@ -91,6 +95,83 @@ func (mtv *MessagesTextView) onInputCapture(e *tcell.EventKey) *tcell.EventKey {
...
+					return nil
+				}
+			}
+		}
+		return nil
+	case mtv.app.Config.Keybindings.DownloadAttachment:
+		if ms[mtv.app.SelectedMessage].Attachments != nil {
+			// Download the files to the configured location
+			for _, a := range ms[mtv.app.SelectedMessage].Attachments {
+				f, err := os.Create(mtv.app.Config.General.DownloadLocation + a.Filename)
```
ghost: You can use `filepath.Join()` to create an OS-specific separator-joined path.

ghost (approved):

ghost:
Thank you for your contribution!

peepo5:
Hi, this does not seem to work (at least for me) anymore. Does this work for anyone?

ghost:
> Hi, this does not seem to work (at least for me) anymore. Does this work for anyone?

Hello there! Could you open a separate issue with details about your operating system and what exactly does not work for you?

PR #149, opened 2022-03-18
feat: add message delete and edit

nouun:
Add an option to delete and edit user messages, small bug that messages don't update visually but have confirmed they update in the official client.


ghost:
To update the messages in the `TextView` widget, the messages need to be redrawn to the widget.

PR #161, opened 2022-06-29
feat: add vi-like keybindings for guilds list

matheusfillipe:
All the other tview widgets used on the main app support vi like keybindings by default, like hjkl G and g, and would be nice to have these on the Guild list as well so that vimmers like me would be happy :)

### DIFF
--- a/ui/app.go
+++ b/ui/app.go
@@ -71,29 +71,16 @@ func (app *App) onInputCapture(e *tcell.EventKey) *tcell.EventKey {
 
 	if app.MainFlex.GetItemCount() != 0 {
 		switch e.Name() {
-		// Avoid focusing what is already focused so that the widgets can use shortcuts
 		case app.Config.Keys.ToggleGuildsList:
-			if app.GuildsList.HasFocus() {
-				return e
-			}
 			app.SetFocus(app.GuildsList)
 			return nil
 		case app.Config.Keys.ToggleChannelsTreeView:
-			if app.ChannelsTreeView.HasFocus() {
-				return e
-			}
 			app.SetFocus(app.ChannelsTreeView)
 			return nil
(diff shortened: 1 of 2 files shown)


ui/app.go
```diff
@@ -71,16 +71,29 @@ func (app *App) onInputCapture(e *tcell.EventKey) *tcell.EventKey {
...
 		case app.Config.Keys.ToggleMessagesTextView:
+			if app.ChannelsTreeView.HasFocus() {
+				return e
+			}
 			app.SetFocus(app.MessagesTextView)
 			return nil
 		case app.Config.Keys.ToggleMessageInputField:
+			if app.MessageInputField.HasFocus() {
+				return e
+			}
```
ghost: I think these if-statements are unnecessarily making the code longer and ugly.

ui/guilds.go
```diff
@@ -26,9 +27,31 @@ func NewGuildsList(app *App) *GuildsList {
 	gl.SetBorder(true)
 	gl.SetBorderPadding(0, 0, 1, 1)
 	gl.SetSelectedFunc(gl.onSelected)
+	gl.SetInputCapture(gl.onInputCapture)
 	return gl
 }
 
+func (gl *GuildsList) onInputCapture(e *tcell.EventKey) *tcell.EventKey {
+	var item = gl.List.GetCurrentItem()
```
ghost: Also, you can use the short variable declaration operator (:=) here:
```go
i := gl.List.GetCurrentItem()
```

ui/app.go
matheusfillipe: Well I thought on these ifs because the default "g" for guild conflicts with the vi-like vi to go up, but i guess the right solution is have the user remap it anyway.

PR #170, opened 2022-08-11
fix: MFA form not set as application root

### LINKED ISSUE #158: MFA not working
I'm running the current `master` (`d5c35ec` at the time of writing this), built through AUR.  
When I try to log in I first get the expected `Email` and `Password` fields. Pressing the `Login` button makes the UI unresponsive for a fraction of a second. I know this because pressing it many times consecutively queues up presses and so increases the unresponsive duration.  
From reading the code I would expect the login window to be re-rendered with an MFA token field, but that never happens. I don't get a `panic` either, which leads me to believe that the login attempt doesn't produce an error.
I don't see any hint of a `--debug` flag or something similar, so I'm a bit at a loss for how to investigate further, short of getting started with golang and running discordo in the debugger.

ghost:
Closes #158

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ Discordo is a lightweight, secure, and feature-rich Discord terminal client. Hea
 
 ## Installation
 
-## Prebuilt binaries
+### Prebuilt binaries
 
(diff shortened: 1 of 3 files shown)


AkechiShiro:
Hey I'm getting MFA triggered even if I don't have it enabled on my account, is that intended ?

Issue #165, opened 2022-07-19
Nothing happens after pressing "LOGIN" button.


Hexzatite:
Title is what it says. Even if I put in a knowingly fake email and password, nothing happens when I try and press the login button. It hangs for a split second, then does nothing. No errors that I can visually see however.
Would be helpful if I knew the location of some sort of log file or something of the sort, but I can't find any, apologies for that.

OS: Arch 5.18.12-arch1-1 (Archcraft)
Terminal: Alacritty 0.10.1-2
Discordo: r119.797d557-1 (Version currently on the AUR)

gohoski:
You can try logging in using a token. This is probably the only way that is working

Hexzatite:
> You can try logging in using a token. This is probably the only way that is working

I could try, How would I go about doing that? Getting the token off the web version and then passing it to discordo?

ghost:
@Ahndrek-LiCyri, thanks for the bug report, I will look into this issue.

> I could try, How would I go about doing that? Getting the token off the web version and then passing it to discordo?

You can login using the client authentication token by supplying it to the `token` flag. Eg: `discordo --token "TOKEN HERE"`

hajosattila:
Same problem here...

Arch 5.18.14-arch1-1
Sway 1.7
Alacritty 0.10.1
Discordo r447.205c9da-1

discordo --token "TOKEN HERE" working fine, but if I put the token in config.json, it doesn't work.

Sorry for my bad English

ghost:
Discord sends an email to verify login if the user has logged in from a new location; go to your mail inbox and verify the login request to continue.

hajosattila:
> Discord sends an email to verify login if the user has logged in from a new location; go to your mail inbox and verify the login request to continue.

I have not received an e-mail.

markaudacity:
Same issue here.

PR #209, opened 2022-09-07
fix: re-add `keyring.Set` call

ghost:
was accidentally removed in #202


ghost:
(apologies for the short PR, I didn't know where to reach out about it)

main.go
```diff
@@ -80,6 +80,8 @@ func main() {
 				log.Fatal(err)
 			}
 
+			keyring.Set(name, "token", tkn)
```
ghost: Can you goroutinize this?

```suggestion
			go keyring.Set(name, "token", tkn)
```

ghost: It could be my system, but I had that initially and it never set the token, and also froze the UI

ghost: Possibly- it's what happened in #165, I remember facing it too

ghost: This is not the reason for issue #165, it has something to do with captcha.

ghost: Okay, my bad. After a few more attempts it seems it was just logging in that was taking really long, but it was coincidentally faster with this change at some point.

ghost (approved):
Thanks for contributing!

PR #211, opened 2022-09-07
Build via Docker and install in Makefile

replicant0wnz:
Allow building via Docker in order to by-pass installing go. Also added `install` to `Makefile`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,3 +17,7 @@ fmt:
 .PHONY: clean
 clean:
 	go clean
+
+.PHONY: install
+install:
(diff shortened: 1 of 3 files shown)


ghost:
You can download the [pre-built binaries](https://github.com/ayntgl/discordo#prebuilt-binaries) that are built on push to default branch using the CI to avoid installing Go and building it from source.

replicant0wnz:
Well, thought I'd push it up as I build via source and had the `Makefile` handy ;-)

PR #213, opened 2022-09-08
docs(README): add FreeBSD in Package manager section

lcook:
I've committed `discordo` to the FreeBSD ports repository in [4747d3fc](https://cgit.freebsd.org/ports/commit/?id=4747d3fcdd2520f83b1274ba1e25087f031d4371), so added the corresponding instructions to install through either the FreeBSD package manager or ports collection in the README.


ghost (approved):
Thanks for submitting! :heart:

PR #228, opened 2022-10-06
Use group chat name, with join usernames as a fallback only!

ShadiestGoat:
The group chat name is sometimes available, why not use it?


ghost (approved):
LGTM!

PR #229, opened 2022-10-06
Fix scroll

ShadiestGoat:
This PR fixes the message scroll. Before, it used to panic whenever it arrow keys were pressed

It removes the `MessageView.selectedMessage`, which while being an efficient method in theory, caused problems since it wasn't set anywhere. 
The new method uses `selInd` (selected Index) as a local variable, which is fetched from the highlighted region's ID. Admittedly, this means it loops over all the messages every time, but I don't see it as a problem since its not done over an array of 100000 elements. I believe that maximum that this project stores is about 100.

### DIFF
--- a/ui/actions_view.go
+++ b/ui/actions_view.go
@@ -96,12 +96,6 @@ func (v *ActionsView) mentionReplyAction() {
 }
 
 func (v *ActionsView) selectReplyAction() {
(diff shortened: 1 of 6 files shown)


ghost:
In earlier versions, it used to use GetHighlights method instead of storing indexnof the element in the slice; however, tjis was noticeably very slow in active channels. Regarding the limit, the message create event appends the new message to the slice, so the limit does not really exist.

PR #236, opened 2022-10-06
Fix reply

ShadiestGoat:
This should fix the reply in DMs & other places. 

Before the ref. was the message being replied to by the message that the user is replying to. Now its the correct one.

Side note, the git is like very messed up at this point, and I think there are changes unrelated to this pr (see line 38-47 util.go, its from the first PR i did, which was merged)


ShadiestGoat:
Don't know why the checks here failed - it seems to build fine on my machine (linux, arch btw)

ghost:
Please fix the commit history and open a pull request again.

PR #237, opened 2022-10-06
Fix the reply

ShadiestGoat:
This fixes the reply with the same method as is done in #236, but with a good commit history


ghost (approved):
Thanks!

PR #238, opened 2022-10-06
Fix scroll

ShadiestGoat:
Fixes scroll using region using `v.selectedMessage` (set in `SetHighlightedFunc`)

### DIFF
--- a/ui/messages_view.go
+++ b/ui/messages_view.go
@@ -28,6 +28,16 @@ func newMessagesView(c *Core) *MessagesView {
 		v.core.App.Draw()
 	})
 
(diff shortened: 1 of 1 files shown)


ghost:
This method is relatively very inefficient for large channels since you are iterating through the slice of messages every time a new region is highlighted.

ShadiestGoat:
> This method is relatively very inefficient for large channels since you are iterating through the slice of messages every time a new region is highlighted.

This happens anyways, since it is being redrawn. This is the method that is most effective to have all kinds of features work properly. The solution to large channels is not to remove anything that iterates over the messages but to limit the cached messages to something more reasonable - around 300-500.

PR #239, opened 2022-10-06
Fix perms in action menu

ShadiestGoat:
In PR #237 I forgot to add an author check for the delete button. This PR not just remedies this, but also adds a perm check only if this is not happening in a DMish channel.


ghost (approved):
Thank you!

PR #243, opened 2022-10-15
Use channel ID to sort private channels by default

ShadiestGoat:
This will fix comments like [this](https://github.com/ayntgl/discordo/issues/230#issuecomment-1276807153). It puts the empty channels into the correct position


ghost (approved):
LGTM!

PR #292, opened 2023-04-28
feat: add config flag to specify different path

Ohkthx:
This will allow users to specify a different location for **config.yml** besides what is returned by `os.UserConfigDir()`.  Users whose OSs do not default to `$XDG_CONFIG_HOME` which typically resides at `$HOME/.config`, will be able to override the default configuration location to adhere to those standard practices.


ayn2op:
If `XDG_CONFIG_HOME` is either empty or not set, it defaults to `$HOME/.config`.

internal/config/config.go
```diff
@@ -37,20 +38,52 @@ func defConfig() Config {
 	}
 }
 
-func Load() error {
+func getPath(optionalPath string) (string, error) {
+	// Trigger an error if config flag used but is empty.
+	if len(optionalPath) == 0 {
```
ayn2op: Strings are comparable in Go.

```suggestion
	if optionalPath == "" {
```

ayn2op (approved):
LGTM! 🚀

God-damnit-all:
I'm not sure if it's more appropriate to talk about this in here or the commit itself, but this as well as logPath were both removed last week without any explanation, I was using both of them to move them to the same directory I launch discordo from.

Why did you remove this?

PR #296, opened 2023-05-09
ctrl + g to toggle guilds display

### LINKED ISSUE #294: feat: show/hide guilds tree widget
When writing in a Direct Message with a friend, i have no use to see all the other DMs, so I would like to hide that. Would this be possible? Maybe it can be done via a setting in the config file?

jeffparla:
resolves #294 
wish tview had hiding of certain elements, but this is a hacky fix that works well enough

### DIFF
--- a/main_flex.go
+++ b/main_flex.go
@@ -6,8 +6,6 @@ import (
 	"github.com/rivo/tview"
 )
 
-var guildsVisible bool = true
-
 type MainFlex struct {
 	*tview.Flex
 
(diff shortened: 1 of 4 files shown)


main_flex.go
```diff
@@ -6,9 +6,12 @@ import (
 	"github.com/rivo/tview"
 )
 
+var guildsVisible bool = true
```
ayn2op: This is not necessary, you can easily check whether the guilds tree is visible or not using `(*tview.Flex).GetItem` or `(*tview.Flex).GetItemCount`.

jeffparla:
solved in #297 :roll_eyes:

PR #300, opened 2023-05-25
Added ability to change reply indicator

ghost:
And added ability to change the place of timestamp. Also fixed a bug when the user tries to color the background of the author name or an attribute.

First time using Go. If there is anything I should change, I will be happy to hear about it.

Here is a display of my setup (bright purple with timestamps on the left and padded with spaces when needed)
image


ayn2op:
Please create separate PRs instead, everything else looks fine.

ghost:
OK. gonna make 2 PRs, one for reply indicator and another one for the timestamp and the bug.

PR #301, opened 2023-05-26
Added ability to change reply indicator

ghost:


messages_text.go
```diff
@@ -86,7 +82,11 @@ func (mt *MessagesText) createMessage(m discord.Message) {
 	}
 }
 
-func (mt *MessagesText) createHeader(w io.Writer, m discord.Message) {
+func (mt *MessagesText) createHeader(w io.Writer, m discord.Message, isreply bool) {
+	if isreply {
+		fmt.Fprintf(mt, "[::d]%s", config.Current.Theme.MessagesText.ReplyIndicator)
+	}
+
```
ayn2op: The `isreply` parameter can be replaced with `m.ReferencedMessage != nil` in the `createHeader` function.

ghost: Huh, never thought of that.

ghost: Well, no it dose not work.
A reply message -> dose have ReferencedMessage
ReferencedMessage is passed -> dose not have ReferencedMessage
Then the message is passed -> dose have ReferencedMessage
If we used
```go
if m.ReferencedMessage != nil {
    fmt.Fprintf(mt, "[::d]%s", config.Current.Theme.MessagesText.ReplyIndicator)
}
```
Then the following will be printed:
```
User1 This is a message
User1 This is a message
╭User2 This is a reply to that message
```
instead of:
```
User1 This is a message
╭User1 This is a message
User2 This is a reply to that message
```

messages_text.go
```diff
@@ -86,7 +82,11 @@ func (mt *MessagesText) createMessage(m discord.Message) {
 	}
 }
 
-func (mt *MessagesText) createHeader(w io.Writer, m discord.Message) {
+func (mt *MessagesText) createHeader(w io.Writer, m discord.Message, isreply bool) {
+	if isreply {
+		fmt.Fprintf(mt, "[::d]%s", config.Current.Theme.MessagesText.ReplyIndicator)
+	}
```
ayn2op:
```suggestion
func (mt *MessagesText) createHeader(w io.Writer, m discord.Message, isReply bool) {
	if isReply {
		fmt.Fprintf(mt, "[::d]%s", config.Current.Theme.MessagesText.ReplyIndicator)
	}
```

PR #302, opened 2023-05-27
Fixed bug where editor wouldn't launch

StandingPadAnimations:
For some reason, maybe a quirk of my terminal or a Go quirk, Discordo couldn't launch an editor. Launching Neovim would hang and Nano would stright up fail. Some poking around found that it was due to Stdout. For some reason, setting Stdout to a string just didn't work.

Now I could try to capture Stdout, but I decided to opt for the simple route of "create a temporary file and extract the message from there".

This is my first time touching Go at all, so it's not the best, but it fixes this issue.


ghost:
Creating `discord_msg.txt` on `/tmp` will make 2 clients interfere with each other when running in the same machine.
Use a unique file name. Something like `discord_msg_$$.txt` where `$$` is the current process ID.
And don't forget to delete it. It's annoying to see 50 or so files in `/tmp`. Some people don't restart their machines for weeks.

message_input.go
```diff
@@ -116,10 +116,26 @@ func (mi *MessageInput) launchEditorAction() {
 	if e == "default" {
 		e = os.Getenv("EDITOR")
 	}
+	
+	// Create a file called discord_msg.txt that we'll
+	// open in the editor. The reason is because capturing
+	// Stdout to a variable actually causes editors to not
+	// work for some reason, so we're going with the more
+	// reliable method.
+	msg_file := os.TempDir() + "/discord_msg.txt"
```
sewnie:
```suggestion
	msg_file := filepath.Join(os.TempDir(), "discord_msg-" + os.Getpid() + ".txt")
```

StandingPadAnimations:
Alright I've now made it so the message file includes the PID of the instance, and I've also added deleting the file whenever any error occurs (with the exception of file creation since I'm not sure if the file would exist when there's an error in creating the file).

sewnie (commented):
Why not defer the `os.Remove(msg_file)` call instead of calling it twice?

StandingPadAnimations:
Sorry about the extra commit. I was working on image viewer support and pushed to the wrong branch.

> Why not defer the `os.Remove(msg_file)` call instead of calling it twice?

Again, this is my first time even touching Go, so stuff like deferring is new to me

message_input.go
ayn2op: https://pkg.go.dev/os#CreateTemp

StandingPadAnimations:
Alright I've added the suggested change to `os.CreateTemp()`. It seems to drastically reduce the amount of code needed for creating the file.

message_input.go
```diff
@@ -116,10 +116,28 @@ func (mi *MessageInput) launchEditorAction() {
 	if e == "default" {
 		e = os.Getenv("EDITOR")
 	}
-
-	cmd := exec.Command(e)
-	var b strings.Builder
-	cmd.Stdout = &b
+	
+	// Create a temporary (with discord_msg inn the name) file
```
ghost:
```suggestion
	// Create a temporary (with discord_msg in the name) file 
```

message_input.go
```diff
@@ -116,10 +116,28 @@ func (mi *MessageInput) launchEditorAction() {
...
-	cmd := exec.Command(e)
-	var b strings.Builder
-	cmd.Stdout = &b
+	
+	// Create a temporary (with discord_msg in the name) file 
+	// that we'll open in the editor. The reason is because 
+	// capturing Stdout to a variable actually causes editors 
+	// to not work for some reason, so we're going with the more
+	// reliable method.
+	f, err := os.CreateTemp("", "discord_msg")
```
ayn2op:
```suggestion
	f, err := os.CreateTemp("", config.Name + "-*.md")
```

ghost:
@StandingPadAnimations Can you change this PR so it can be merged with the new file locations?

StandingPadAnimations:
Sorry for the long wait, I've now updated this branch

ghost:
On 23/07/20 07:56PM, StandingPad wrote:
> Sorry for the long wait, I've now updated this branch
> 
> -- 
> Reply to this email directly or view it on GitHub:
> https://github.com/ayn2op/discordo/pull/302#issuecomment-1644915941
> You are receiving this because you commented.
> 
> Message ID: ***@***.***>

Good.

--

cmd/run/message_input.go
```diff
@@ -111,10 +111,28 @@ func (mi *MessageInput) launchEditorAction() {
...
-	cmd.Stdout = &b
+	
+	// Create a temporary (with discord_msg in the name) file 
+	// that we'll open in the editor. The reason is because 
+	// capturing Stdout to a variable actually causes editors 
+	// to not work for some reason, so we're going with the more
+	// reliable method.
+	f, err := os.CreateTemp("", config.Name + "-*.md")
+	if err != nil {
+		log.Println(err)
```
ayn2op, line 122:
```suggestion
		log.Println(err)
		return
```

cmd/run/message_input.go
```diff
@@ -111,10 +111,28 @@ func (mi *MessageInput) launchEditorAction() {
...
+	}
+	f.Close()
+	
+	// Defer this so the file is deleted when the
+	// function returns, regardless of failiure or not
```
ayn2op: Typo.

```suggestion
	// function returns, regardless of failure or not
```

cmd/run/message_input.go
```diff
@@ -123,6 +141,15 @@ func (mi *MessageInput) launchEditorAction() {
 			return
 		}
 	})
+	
+	// One may ask "Why don't we remove the file earlier?". Well,
+	// the file won't contain any message up until this point, and the file
+	// is created in /tmp anyway so it'll be cleared on a reboot.
+	var msg, read_err = os.ReadFile(f.Name())
```
ayn2op: Use `:=` instead of `var`.

```suggestion
	msg, read_err := os.ReadFile(f.Name())
```

cmd/run/message_input.go
```diff
@@ -123,6 +142,15 @@ func (mi *MessageInput) launchEditorAction() {
 			return
 		}
 	})
+	
+	// One may ask "Why don't we remove the file earlier?". Well,
+	// the file won't contain any message up until this point, and the file
+	// is created in /tmp anyway so it'll be cleared on a reboot.
+	msg, read_err := os.ReadFile(f.Name())
```
ayn2op: Convert `read_err` to camel case.

PR #303, opened 2023-05-27
Added ability to change timestamps location

ghost:
This is a follow up for #301

### DIFF
--- a/cmd/run/messages_text.go
+++ b/cmd/run/messages_text.go
@@ -83,8 +83,10 @@ func (mt *MessagesText) createMessage(m discord.Message) {
 }
 
 func (mt *MessagesText) createHeader(w io.Writer, m discord.Message, isReply bool) {
+	time := m.Timestamp.Format(time.Kitchen)
+
 	if config.Current.Timestamps && config.Current.TimestampsBeforeAuthor {
-		fmt.Fprintf(w, "[::d]%7s[::-] ", m.Timestamp.Format(time.Kitchen))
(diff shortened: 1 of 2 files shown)


messages_text.go
```diff
@@ -83,13 +83,17 @@ func (mt *MessagesText) createMessage(m discord.Message) {
...
 		fmt.Fprintf(mt, "[::d]%s", config.Current.Theme.MessagesText.ReplyIndicator)
 	}
 
-	fmt.Fprintf(w, "[%s]%s[-] ", config.Current.Theme.MessagesText.AuthorColor, m.Author.Username)
+	fmt.Fprintf(w, "[%s]%s[-:-:-] ", config.Current.Theme.MessagesText.AuthorColor, m.Author.Username)
 
-	if config.Current.Timestamps {
+	if config.Current.Timestamps && !config.Current.TimestampsBeforeAuthor {
 		fmt.Fprintf(w, "[::d]%s[::-] ", m.Timestamp.Format(time.Kitchen))
 	}
```
ayn2op: Compute the formatted timestamp if the `timestamps` field is set as `true` in the configuration file and then write the computed timestamp to the `io.Writer` depending on the value of `timestamps_before_author` instead.

ghost: So you mean `m.Timestamp.Format(time.Kitchen)` be used once or what?

ayn2op: Yes.

ghost:
OK. Got it to the new project format.

BTW why you moved to kong for CLI?

PR #304, opened 2023-05-28
Added unread indicators

StandingPadAnimations:
2 PRs in the span of 2 days, holy smokes.

Anyway, yes, this PR adds unread indicators. This is done by iterating over the `ReadStates` in the `ReadyEventExtras` field of `discordState.Ready()`. We do check if the channel IDs match, and if so whether the last message IDs match. If they do, then we simply ignore it. Otherwise, we set a tag. This tag is then combined with the channel names, so unread channels will be italicized and bolded.

There are still issues though:
- `ReadStates` are not updated, even when a message is sent
- Some people might have a hard time to tell the difference between unread and read channels
- This does seem to affect voice channels for some reason, I have no idea why
- No indicator for which messages are unread
- No respect for muted channels

Example of unread channels side-by-side read ones:
image


StandingPadAnimations:
And yes this branch is based off of #302

guilds_tree.go
```diff
@@ -75,6 +75,17 @@ func (gt *GuildsTree) createGuildNode(n *tview.TreeNode, g discord.Guild) {
...
+	tag := "[::d]"
+	read_states := discordState.Ready().ReadyEventExtras.ReadStates
+
+	for i := range read_states {
+		if read_states[i].ChannelID == c.ID {
+			if read_states[i].LastMessageID != c.LastMessageID {
+				tag = "[::bi]"
+			}
+		}
+	}
```
ghost: You could do
```go
if read_states[i].ChannelID == c.ID && read_states[i].LastMessageID != c.LastMessageID {
	tag = "[::bi]"
}
```
instead of:
```go
if read_states[i].ChannelID == c.ID {
	if read_states[i].LastMessageID != c.LastMessageID {
		tag = "[::bi]"
	}
}
```

guilds_tree.go
```diff
@@ -104,7 +115,7 @@ func (gt *GuildsTree) channelToString(c discord.Channel) string {
 		s = c.Name
 	}
 
-	return s
+	return tag + s + "[-]"
```
ghost: Use "[::-]" instead of "[-]".
Because "[::-]" will reset only the attributes.

ghost:
Why not make the `tag` configurable?
Add `guild_unread_color` and `guild_read_color` in `internal/config/theme.go` under the `GuildsTreeTheme` structure.
Then make the default for read as `::d` and for unread as `::bi`.

StandingPadAnimations:
> Why not make the `tag` configurable? Add `guild_unread_color` and `guild_read_color` in `internal/config/theme.go` under the `GuildsTreeTheme` structure. Then make the default for read as `::d` and for unread as `::bi`.

Implemented in 2655fc5

StandingPadAnimations:
Alright I've now added guild indicators, in addition to indicators for pings.

Should we make guild indicators togglable? They make guilds take additional time to load (since there's 5 for loops at max, due to some crazy aspects of `ReadyEventExtras`

StandingPadAnimations:
Ok, now channels will be marked as read when you open them. Of course, I still need to get this working only on unread channels (and then get the channels to be updated in the UI).

StandingPadAnimations:
Ack is now called only for unread channels (that aren't muted). This is done by checking the beginning tag and seeing if it matches the Unread or Mention indicator.

PR #307, opened 2023-06-04
Add bookmark feature

emanueljg:
This PR implements a very basic bookmark system for discordo. 

A bookmark is a channel (including DMs) that exists top-level, before the DM node and list of servers, and is always reachable through the server list through a keyboard shortcut. Bookmarks are added in the config. Here is an example:
  ```yml
  bookmarks:
      - alias: foobar general
        id: 741347063077535874
      - alias: bazchan
        id: 568306982717751328
  ```
- Besides that, I have also added 2 keyboard shortcuts where the first goes to (not selects!) the first bookmark, and one shortcut that goes to the first non-bookmark node (The DMs in other words). 
- I went through a few different implementations of this idea. I finally settled on bookmarks being primarily being identified by their `ChannelID` and not their name. Not only does this result in cleaner code and seemingly O(1) channel fetching, but I think it guards against the problem of channels switching names.
- I'll wait before I add add my keybinds etc to any documentation as maintainers might suggest changes before they become "written in stone". 

This is not only my first discordo PR but also my first golang code written ever, so I don't know if I've written idiomatic Go or not. I'm open to changes.


ghost:
Would it be possible to add bookmarks and remove them throw the TUI?

It should be simple, like:  
Read the config file, parse it, check for channel ID in the bookmarks list, if the ID exist, delete the bookmark, else add a new bookmark, write the new config file.

emanueljg:
> Would it be possible to add bookmarks and remove them throw the TUI?
> 
> It should be simple, like: Read the config file, parse it, check for channel ID in the bookmarks list, if the ID exist, delete the bookmark, else add a new bookmark, write the new config file.

I'm not sure I understood your explanation of your idea but just the idea of adding/removing bookmarks dynamically with for example a keyboard shortcut sounds interesting, but then I would want the bookmarks to exist in a seperate file. I am strongly opposed to the idea of the program editing user configuration.

I'll probably be too busy to implement this myself the following days as I don't require this quirk myself and I'm already satisfied with this PR as-is, but if anyone wants to add this, feel free.

ghost:
> > Would it be possible to add bookmarks and remove them throw the TUI?
> > It should be simple, like: Read the config file, parse it, check for channel ID in the bookmarks list, if the ID exist, delete the bookmark, else add a new bookmark, write the new config file.
> 
> I'm not sure I understood your explanation of your idea but just the idea of adding/removing bookmarks dynamically with for example a keyboard shortcut sounds interesting, but then I would want the bookmarks to exist in a seperate file. I am strongly opposed to the idea of the program editing user configuration.
> 
> I'll probably be too busy to implement this myself the following days as I don't require this quirk myself and I'm already satisfied with this PR as-is, but if anyone wants to add this, feel free.

I should have explained it better, but your idea is good.
Having a file (e.g. `$HOME/.config/discord/bookmarks`) in a line-terminated format that contain bookmarks is better than trying to change user's `config` file.
Here is an example prototype:
```sh
# channel_ID Alias name
6969420420420 The best channel
1234567890887 Some other channel
```
I will try to implement it my self.
If you didn't see any PRs from my about this, then I failed horribly.

PR #311, opened 2023-06-24
Update configuration.md

mrpink32:
Added toggle binding to docs.
Replaced "Reply" with "Select" for moving up through messages.


ayn2op (approved):
LGTM 🚀

PR #314, opened 2023-07-05
feat: implement editing and deleting messages & s-substitution

sewnie:
applies on-top of #313


sewnie:
note: window is not re-drawn

cmd/run/message_input.go
```diff
@@ -68,35 +68,64 @@ func (mi *MessageInput) sendAction() {
 		return
 	}
 
+	title := mi.GetTitle()
```
ayn2op: Move `title` to closest to where it is used.

cmd/run/message_input.go
```diff
@@ -68,58 +68,75 @@ func (mi *MessageInput) sendAction() {
...
+		} else if strings.HasPrefix(title, "Editing") || substitution {
+			m, err := discordState.EditText(mainFlex.guildsTree.selectedChannelID, ms[mainFlex.messagesText.selectedMessage].ID, text)
+			if err != nil {
+				log.Println(err)
+				return
+			}
+
+			if err := discordState.MessageSet(m, true); err != nil {
+				log.Println(err)
+			}
```
ayn2op:
```suggestion
			err = discordState.MessageSet(m, true)
			if err != nil {
				log.Println(err)
				return
			}
```

ui/login_form.go
```diff
@@ -13,12 +13,14 @@ import (
...
 	*tview.Form
 	Token chan string
+	Error chan error
 }
 
 func NewLoginForm() *LoginForm {
 	lf := &LoginForm{
 		Form:  tview.NewForm(),
 		Token: make(chan string, 1),
+		Error: make(chan error, 0),
```
ayn2op: Channels are unbuffered by default. `0` is redundant here.
```suggestion
		Error: make(chan error),
```

ui/login_form.go
```diff
@@ -13,12 +13,14 @@ import (
 type LoginForm struct {
 	*tview.Form
 	Token chan string
+	Error chan error
```
ayn2op, line 16: This is not related to the PR, please remove it and open a new PR.

PR #327, opened 2023-10-27
implement Cancel of Reply and Message Selection

SRVallee:
Currently, once you have selected a message, there is no option to 'deselect' a message. Additionally, you implicitly reply to any selected message.

Changes:
- Replies are now only explicit. You must press either 'reply' or 'select_reply' keys to reply
- Replies can be cancelled by pressing the 'cancel' key.
- You can deselect a message by pressing the 'cancel' key


ayn2op:
Stale

PR #330, opened 2023-12-05
feat: implement vi-like interface

### LINKED ISSUE #282: Feature Request: Vi-alike Interface
What the title really means is to basically, instead of relying on another vocabulary of combinations of several hotkeys (which are mostly not at all cozy to press; for example, for the CTRL key, you'd need to put some effort to get back to a state when you could type again), maybe just introduce a vi-alike interface. 

Right now pressing ESC clears the input, however the same thing, on my machine is also achieved by pressing CTRL + U; so that's a duplicated hotkey. What we can do instead is that, when we press ESC, it'd put a hint in the input field that we are currently in "Normal Mode" and then make use of all the keys which are easily accessible to control the UI elements, and when we type `:` in the normal, we will be greeted with a command line (we can just use the input field for this) where we can type ":insert" to go back to "Insert Mode".

The benefits of this implementation:
- Super easy and cozy keyboard control of the UI.
- Very easy to use "plugins" that you have planned (like a plugin to find servers and messages stored in cache; another to view avatars of servers and users; another to download the hinted image and convert it to a ASCII and then display it on the screen, this can be achieved by Go libraries like [this](https://github.com/qeesung/image2ascii#library-usage); another to use `jj` for QWERTY users, or `nn` for Colemak users to activate "Normal Mode" in "Insert Mode").
- Very easy to define custom commands.

ayn2op:
Please submit feedback here.

Normal mode:
image

Insert mode:
image

Closes #282

### DIFF
--- a/cmd/attachment_image.go
+++ b/cmd/attachment_image.go
@@ -47,11 +47,11 @@ func newAttachmentImage(a discord.Attachment) (*AttachmentImage, error) {
 }
 
 func (ai *AttachmentImage) onInputCapture(event *tcell.EventKey) *tcell.EventKey {
-	if event.Name() == cfg.Keys.Cancel {
-		app.SetRoot(mainFlex, true)
-		app.SetFocus(mainFlex.messagesText)
-		return nil
-	}
+	// if event.Name() == cfg.Keys.Cancel {
+	// 	app.SetRoot(mainFlex, true)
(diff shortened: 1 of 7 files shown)


ghost:
What if the user didn't want Vi-like keybindings?

ayn2op:
> What if the user didn't want Vi-like keybindings?

A configuration field can be set to `false` to turn off vi interface.

Issue #329, opened 2023-11-27
Show other usernames in different colors


MysticAx0lotl:
Seeing everyone's username in the same color is really confusing. Is it possible to add some kind of delimiter so that everyone's usernames don't look identical?

PR #347, opened 2024-02-26
Add different color for own messages

cyberme0w:
I saw a similar feature requested in #329. For me, it would be enough to see my own messages, as it more easily allows me to check if there's something new. I didn't immediately find a way
to get the username from state, so I added it as a settings entry.

I didn't add to the documentation yet, because I would like to wait for feedback - if this sounds good, I'll make sure to update it.


ayn2op:
Names will be highlighted using the highest role in the future.

cyberme0w:
Ah, I see. That's good to hear. I'll go ahead and close the PR.

PR #353, opened 2024-03-01
Draft: Add datetime of original message to title when replying

cyberme0w:
Hi!

It's not always obvious which message is actually being replied to, since the title is only set to the username that sent the original message. With this PR, a timestamp (time for same day, otherwise datetime) is added afterwards, turning this:

```
╔Replying to cyberme0w════════════════════════════╗
║[message input here]                             ║
╚═════════════════════════════════════════════════╝
```

into this (original message from same day as reply):

```
╔Replying to cyberme0w @ 15:45:32═════════════════╗
║[message input here]                             ║
╚═════════════════════════════════════════════════╝
```

and this (original message from previous day or older):

```
╔Replying to cyberme0w @ 2024-01-01 15:45:32══════╗
║[message input here]                             ║
╚═════════════════════════════════════════════════╝
```

If this sounds interesting, I can also add support for the following:
- config entry to enable/disable timestamps in reply title
- configurable timestamp format string, or perhaps a selection of typical datetime formats 

**Note:** there is currently a bug with replies - when replying to a message, if the highlight is moved, the message being replied to will be the highlight *instead of the message that was highlighted when pressing the reply shortcut*. I have a fix for that which I'll add in a separate PR. This means that, while that bug exists, the timestamp added in this PR might not match the message being replied to.

Cheers!

### DIFF
--- a/cmd/messages_text.go
+++ b/cmd/messages_text.go
@@ -193,9 +193,9 @@ func (mt *MessagesText) replyAction(mention bool) {
 	sy, sm, sd := ms[mt.selectedMessage].Timestamp.Time().Date()
 	ny, nm, nd := time.Now().Date()
 	if sy == ny && sm == nm && sd == nd {
-		timestamp = ms[mt.selectedMessage].Timestamp.Time().Format("15:04:05")
+		timestamp = ms[mt.selectedMessage].Timestamp.Time().Format(cfg.TimestampsFormatTime)
 	} else {
-		timestamp = ms[mt.selectedMessage].Timestamp.Time().Format("2006-01-02 15:04:05")
+		timestamp = ms[mt.selectedMessage].Timestamp.Time().Format(cfg.TimestampsFormatDatetime)
 	}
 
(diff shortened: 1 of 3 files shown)


ghost:
I think
```
╔Replying to cyberme0w @ 15:45:32 2024-01-01══════╗
║[message input here]                             ║
╚═════════════════════════════════════════════════╝
```
is better because how the eyes works when looking at something seemlier.

ghost:
Or have a configuration option

cyberme0w:
I'll set up the config option and add your format as the default to the config, along with a comment explaining what it is. I think it also makes sense to have a 'time' and a 'date' string, as that makes generating the 'same-day' vs. 'different-day' timestamps easier.

cyberme0w:
Perhaps it also makes sense to set the config option as generic timestamp formats - that way they could be reused to format the individual messages' timestamps in Messages Text.

cyberme0w:
Closing, as I will no longer be working on this PR.

Issue #348, opened 2024-02-26
Replace IDs in mentions with usernames


cyberme0w:
Hi!

Currently, discordMessage.Content displays mentions with USERID, a number with which the user can not really understand who was mentioned.

I would like to propose to replace it with the user's actual username, so that

```
User: I'm going to ping <@123123123>
AnotherUser: Thanks for the ping!
```

becomes 

```
User: I'm going to ping @AnotherUser
AnotherUser: Thanks for the ping!
```

Additionally, to make mentions more visible, I would suggest making them bold and/or underlined, possibly making it a config option.

I do have a patch for this already (without the config bit) and could open a PR if there is interest.
Cheers!

PR #356, opened 2024-03-03
Replace user IDs with usernames in Mentions

cyberme0w:
Hi! I wrote a small PR for the feature I proposed in #348. I think the current format (bold+underlined) should be visible enough to stand out.

Cheers!

### DIFF
--- a/cmd/messages_text.go
+++ b/cmd/messages_text.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"strings"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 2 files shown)


cmd/messages_text.go
```diff
@@ -111,8 +112,24 @@ func (mt *MessagesText) createHeader(w io.Writer, m discord.Message, isReply boo
...
+func parseIDsToUsernames(m discord.Message) string {
+	c := m.Content
+	for _, mention := range m.Mentions {
+		mentionUserID := fmt.Sprintf("<@%s>", mention.User.ID.String())
+		mentionUsername := fmt.Sprintf("__**@%s**__", mention.User.Username)
+		c = strings.Replace(c, mentionUserID, mentionUsername, -1)
+	}
+	return c
+}
+
```
ayn2op: Just use a [strings.Replacer](https://pkg.go.dev/strings#example-NewReplacer) here.

cmd/messages_text.go
```diff
@@ -111,8 +112,28 @@ func (mt *MessagesText) createHeader(w io.Writer, m discord.Message, isReply boo
...
+	var toReplace []string
+
+	for _, mention := range m.Mentions {
+		toReplace = append(toReplace,
+			fmt.Sprintf("<@%s>", mention.User.ID.String()),
+			fmt.Sprintf("__**@%s**__", mention.User.Username))
+	}
+	r := strings.NewReplacer(toReplace...)
+	c = r.Replace(c)
+	return c
```
ayn2op:
```suggestion
	var toReplace []string
	for _, mention := range m.Mentions {
		toReplace = append(toReplace,
			fmt.Sprintf("<@%s>", mention.User.ID.String()),
			fmt.Sprintf("__**@%s**__", mention.User.Username),
		)
	}

	return strings.NewReplacer(toReplace...).Replace(m.Content)
```

cmd/messages_text.go
```diff
@@ -111,8 +112,28 @@ func (mt *MessagesText) createHeader(w io.Writer, m discord.Message, isReply boo
...
+}
+
 func (mt *MessagesText) createBody(w io.Writer, m discord.Message) {
-	fmt.Fprint(w, markdown.Parse(tview.Escape(m.Content)))
+	var body string
+	if len(m.Mentions) > 0 {
+		body = parseIDsToUsernames(m)
+	} else {
+		body = m.Content
+	}
```
ayn2op, line 133:
```suggestion
	}

```

cmd/messages_text.go
cyberme0w: Maybe it would make sense to return the string-pairs for the replacer, instead of replacing it inside the function. So something like:

```
func getReplacerStringsForUsername() {
  ...
  return toReplace
}
```

and then replace the strings inside `createBody()`.

ayn2op (approved):

Issue #355, opened 2024-03-02
Dim referred message's content when displaying a message reply


lukalotl:
Replies could be made more readable / clarify that the replied message is not new by dimming them to a darker color than the default white, similar to how the official client dims its referenced replies.

PR #357, opened 2024-03-03
Dim body of reply messages

cyberme0w:
I saw #355 and agree that it would make it easier to distinguish between regular messages and messages in the header of replies.


ayn2op (approved):
LGTM!

lukalotl:
Thanks for working on this cyberme0w!

Issue #361, opened 2024-03-08
Timestamps don't show in local time


10kmotorola:
Date is set correctly on my system:
```
$ date
Fri Mar  8 12:24:46 PM EST 2024
```
Yet timestamps are in UTC. I'm using Hyprland fwiw.

cyberme0w:
Looks like something where
https://pkg.go.dev/time#ParseInLocation and https://pkg.go.dev/time#Location would help.

Edit: also somewhat related, the timestamps are always time-only, even for messages older than a day, which is a bit confusing. It'd be nice to be able to see the date too for those older messages.

PR #362, opened 2024-03-10
Add timezone info to message timestamps

cyberme0w:
Hi! This is a fix for #361 which adds timezone information to message timestamps, allowing them to display the correct time (as opposed to hard-coded UTC).

It is important to note that this does not currently work on Android/Termux due to https://github.com/termux/termux-packages/issues/2622 (or more specifically https://github.com/golang/go/issues/20455).

If you think it's useful, I can also look into adding a config option to hard-code the timezone as a workaround for those users.

Cheers!


ayn2op (approved):
LGTM! Thank you for the PR! ❤️

PR #363, opened 2024-03-12
Fix issues when swapping between Message Input and EDITOR

cyberme0w:
Hi!

This PR fixes two issues I noticed when swapping between the Message Input and EDITOR:
1. The EDITOR is initialized empty (CreateTemp generates an empty file), even if the user had already typed something.
2. When returning from the EDITOR, a newline is appended for some reason (tested with neovim and nano)

Issue 1 is solved by writing the contents of the Message Input to the temp file.
Issue 2 is solved by trimming leading and trailing whitespace, as Discord already does that after the message is sent.

Cheers!

### DIFF
--- a/cmd/message_input.go
+++ b/cmd/message_input.go
@@ -120,6 +120,7 @@ func (mi *MessageInput) launchEditorAction() {
 		log.Println(err)
 		return
 	}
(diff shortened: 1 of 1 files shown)


ayn2op:
Thank you!

PR #364, opened 2024-03-13
Feat: Add User List

### LINKED ISSUE #129: Feature request: List and tag users in a channel
Would be great to have the functionality of listing users in a channel and tagging @<name> them in chat.

cyberme0w:
Hi!

This PR revolves around adding a new UI element containing a list of members of the current channel, along with a drop-down menu with commands when selecting a user.

The UserList is built similarly to the GuildsTree. Each user is a node with a UserID reference while commands are children (with the UserID referenced as a string, to be able to distinguish between a user and a command).

This opens up the possibility to add more user-related commands in the future. Currently, I have only implemented appending a mention to the current message based on the highlighted user.

I would like to keep this as a draft for now, as I'm still working out some issues here and there, but would like to know what you think / if you have any suggestions.

Preview:
file

Cheers!

Closes #129

### DIFF
--- a/cmd/guilds_tree.go
+++ b/cmd/guilds_tree.go
@@ -201,6 +201,8 @@ func (gt *GuildsTree) onSelected(n *tview.TreeNode) {
 
 		gt.selectedChannelID = ref
 		app.SetFocus(mainFlex.messageInput)
+
+		mainFlex.userList.updateUsersFromGuildID(c.GuildID)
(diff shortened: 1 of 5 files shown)


cyberme0w:
This would also close #129

cyberme0w:
Closing, as I will no longer be working on this PR.

PR #365, opened 2024-03-13
Move default configuration into config.go

cyberme0w:
Hi!

When adding new features, it is often the case that the config struct is extended. Since the struct was previously generated via the config *file*, which is maintained by the user, it would require them to actively add the missing entries in their configs.

For instance, the new timestamp formats are stored in the config. A user without those entries in their config will see something like:
`-- Replying to [username] @ --`

instead of the expected
`-- Replying to [username] @ 10:45 --`

With a default configuration built into the application:
1. the user can keep their current config file (even though some entries might be the same as the default settings)
2. new features do not require the user to add those entries to the config
3. by default, no actual config file is *required* - making the effort of maintaining a config file optional if the user has no special config

Cheers!

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -151,8 +151,7 @@ func Load() (*Config, error) {
 	cfg.Theme.MessagesText.ReplyIndicator = "╭"
 
 	// Overwrite default values via config file
-	decoder := yaml.NewDecoder(reader)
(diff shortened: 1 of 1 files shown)


ghost:
This is actually a very good feature.

cyberme0w:
I'm going to go ahead and close this, since it's been implemented in #372

Issue #366, opened 2024-03-16
Collapse guild folders by default


epetousis:
Currently I have a whole bunch of guilds thrown into folders, and I only actually have like 5 guilds at the top level. Is there any way to start Discordo with all those folders closed? I couldn't see anything helpful in the tview default keybinds.

PR #367, opened 2024-03-16
Add config option to disable auto-expanding folders

cyberme0w:
Hi!

I personally don't use folders, so I hadn't noticed that folders are automatically expanded, but I saw #366 and agree that the user should at least be able to choose (depending on how many folders you have / servers you've joined it might become very cluttered). In any case, I added a config option that allows the user to choose if they want to expand them or not.

Cheers!


ayn2op (approved):
LGTM! 🚀

epetousis:
Crazy fast turnaround, thank you so much!

Issue #368, opened 2024-03-17
Channel categories without "view channel" permissions are hidden, even if there are viewable channels inside


epetousis:
I'm in a Discord server for uni, and they have a system wherein you select a role, and you are granted access to channels for specific units. Rather than having all these units visible at the top-level, they organise these channels into categories based on the level of unit. They implement this by disabling "view channels" permissions on the channel category (so that you don't see an absolute ton of unrelated channels inside the category), but enabling explicit view permissions on the channels inside depending on which role you select. This works great, but Discordo doesn't support this, [as it will not create the category node due to the lack of "view channel" permissions.](https://github.com/ayn2op/discordo/blob/509f276c886c5463558e31ee9ad7adacbcdac134/cmd/guilds_tree.go#L118)

Discord's behaviour appears to be to show the channel category (and bypass the lack of category view perms) if there are viewable channels inside.

My hack fix is in the attached patch, obviously not a great long-term solution because now a whole bunch of unused categories will show up.

[0001-Show-channel-categories-regardless-of-view-permissio.patch](https://github.com/ayn2op/discordo/files/14625865/0001-Show-channel-categories-regardless-of-view-permissio.patch)

PR #369, opened 2024-03-17
Fix hidden category/channel bug

cyberme0w:
Hi!

I saw #368 and decided to take a bit of a closer look at it. I set up a server like this, which I think emulates the situation described in the issue close enough:
- private-category (visible to admin only)
  - role1 (visible to role1 only)
  - admin (admin only)
- public-category (visible to all)
  - role1 (visible to role1 only)
  - admin (admin only)
- empty-public-category (visible to all)
  - (no channels added here)

Previously, this resulted in the following GuildTree:
- public-category
  - role1
 
With this fix, the GuildTree looks like this for the role1 member:
- private-category
  - role1
- public-category
  - role1

This matches how the web client displays them, too.

The issue was that categories were not being drawn if the user did not have view channel permissions for them. However, a category can be hidden and still have user-visible channels, which should be drawn in the guild tree.

With this fix, the `createChannelNodes()` function now has three steps as follows:
1. Draw all top-level, non-category channels
2. Draw categories that have at least one visible channel (regardless of the categories' own visibility)
3. Draw channels inside (visible) categories if they are visible to the user

Cheers!


ayn2op (approved):
LGTM! <3

epetousis:
Fix definitely works, thanks so much!

PR #375, opened 2024-03-28
Fix TOML configuration in README

cyberme0w:
The example/default configuration in the README is broken and was returning the following error:

`toml: line 13 (last key "keys.normal.guilds_tree"): newlines not allowed within inline tables`


ayn2op (approved):
Thank you! <3

PR #381, opened 2024-04-09
Guildtree Color

AERDU:
This commit adds support for configuring ```tree_color``` which sets the background color for the guildtree's cursor, as well as the text color for the guildtree.

A few things to consider:

1. I'm not quite sure if ```tree_color``` is a name that describes what it does properly.
2. This commit has ```tree_color``` set to ```white``` by default, making the text hard to read. This is the same as the current behavior, so might be a good idea to switch it to ```gray``` or something.
3. It is possible to change ```graphics``` color too. Not sure if I should create a separate option for it, or if ```tree_color``` should apply to it.
4. Preferably the background color and the text color should be separate options, but from my understanding, limitations in tview make the two options the same. I've got very limited knowledge in go, so don't take my word for it.

Either way, this should improve legibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,6 +99,7 @@ title_color = "default"
 background_color = "default"
 
 [theme.guilds_tree]
+tree_color = "white"
 auto_expand_folders = true
 graphics = true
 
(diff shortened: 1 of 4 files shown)


ayn2op:
Stale

PR #430, opened 2024-09-20
feat: render pinned messages

### LINKED ISSUE #429: Not drawing messages with unknown types messes with the next/prev message selection logic
Hi,

the logic for handling cfg.Keys.SelectPrevious/SelectNext assumes every message is actually represented by a line/region. 

https://github.com/ayn2op/discordo/blob/9d001a522e60f3d199b6719ca651a12a089c2538/cmd/messages_text.go#L208

A message like "foo pinned to channel" currently doesn't generate a region because it's of type 6 (discord.ChannelPinnedMessage).

https://github.com/ayn2op/discordo/blob/9d001a522e60f3d199b6719ca651a12a089c2538/cmd/messages_text.go#L86

Easy fix would be a line/region to representing the default case. Stating it's a message with an unknown type.

Best,

EDIT: Better way to solve this would be to rework the selection logic as it doesn't play well with other things either. Like selecting a region with a mouse click.

Retropaint:
Closes #429 

Implemented a basic pinned msg body, along with extra logic to look for m.Reference to jump to that msg (since m.ReferencedMessage is only for replies). This is one half of the issue, letting the 'msg is pinned' actually be seen.

Start and end text regions were put into their own functions, and now cover the entirety of createMessage. This is the second half of the issue, as it encompasses all types of messages. Before this change, blocked messages could not be navigated either (should generally cover other types too, worst come to worst the new funcs will make specific implementation cleaner).


cmd/messages_text.go
```diff
@@ -73,21 +73,33 @@ func (mt *MessagesText) reset() {
...
 			fmt.Fprintln(mt, "[:red:b]Blocked message[:-:-]")
+			mt.endRegion()
 			return
 		}
 	}
 
 	switch m.Type {
+	case discord.ChannelPinnedMessage:
+		fmt.Fprint(mt, "[" + cfg.Theme.MessagesText.PinnedColor + "]" + m.Author.Username + " pinned a message" + "[-:-:-]")
+	break;
```
ayn2op: Redundant `break`

cmd/messages_text.go
```diff
@@ -98,11 +110,11 @@ func (mt *MessagesText) createMessage(m discord.Message) {
 		mt.createHeader(mt, m, false)
 		mt.createBody(mt, m, false)
 		mt.createFooter(mt, m)
-
-		// Tags with no region ID ([""]) don't start new regions. They can therefore be used to mark the end of a region.
-		fmt.Fprint(mt, `[""]`)
+		
 		fmt.Fprintln(mt)
```
ayn2op: Add a default case to the switch block that only displays the name of the author and/or timestamp of the message using the `createHeader` function.

cmd/messages_text.go
```diff
@@ -242,6 +254,13 @@ func (mt *MessagesText) _select(name string) {
 					mt.selectedMessage = i
 				}
 			}
+		} else if ref := ms[mt.selectedMessage].Reference; ref != nil {
+			for i, m := range ms {
+				if ref.MessageID == m.ID {
+					mt.selectedMessage = i
+					break;
+				}
+			}
```
ayn2op: Selecting pinned message should have its own keybinding.

internal/config/theme.go
```diff
@@ -23,6 +23,7 @@ type (
 	MessagesTextTheme struct {
 		AuthorColor     string `toml:"author_color"`
 		EmoteColor      string `toml:"emote_color"`
+		PinnedColor     string `toml:"pinned_color"`
```
ayn2op: Just reuse `ContentColor` configuration field instead of creating a new one.

cmd/messages_text.go
Retropaint: added 'selectPin' keybinding. Let me know if it should be renamed.

cmd/messages_text.go
Retropaint: added the default case, and also moved the Println outside of the switch case so it applies to every message.

cmd/messages_text.go
Retropaint: forgot this wasn't C++, my bad

cmd/messages_text.go
```diff
@@ -80,21 +80,32 @@ func (mt *MessagesText) reset() {
...
+	fmt.Fprint(mt, `[""]`)
+}
+
 func (mt *MessagesText) createMessage(m discord.Message) {
+	mt.startRegion(m)
 	if cfg.HideBlockedUsers {
 		isBlocked := discordState.UserIsBlocked(m.Author.ID)
 		if isBlocked {
 			fmt.Fprintln(mt, "[:red:b]Blocked message[:-:-]")
+			mt.endRegion()
```
ayn2op: Defer the `endRegion` function instead of calling it every time you return from the function:
```go
mt.startRegion(m.ID)
defer mt.endRegion()
```

cmd/messages_text.go
```diff
@@ -80,21 +80,32 @@ func (mt *MessagesText) reset() {
 	mt.Highlight()
 }
 
+// Region tags are square brackets that contain a region ID in double quotes
+// https://pkg.go.dev/github.com/rivo/tview#hdr-Regions_and_Highlights
+func (mt *MessagesText) startRegion(m discord.Message) {
+	fmt.Fprintf(mt, `["%s"]`, m.ID)
+}
```
ayn2op: Accept the ID of the message instead of the entire message:

```suggestion
func (mt *MessagesText) startRegion(msgID discord.MessageID) {
	fmt.Fprintf(mt, `["%s"]`, msgID)
}
```

cmd/messages_text.go
Retropaint: deferred that, and Fprintln as well (blocked msg uses Fprint now)

cmd/messages_text.go
```diff
@@ -80,21 +80,34 @@ func (mt *MessagesText) reset() {
...
+
+// Tags with no region ID ([""]) don't start new regions. They can therefore be used to mark the end of a region.
+func (mt *MessagesText) endRegion() {
+	fmt.Fprint(mt, `[""]`)
+}
+
 func (mt *MessagesText) createMessage(m discord.Message) {
+	mt.startRegion(m.ID)
+	defer fmt.Fprintln(mt)
+	defer mt.endRegion()
```
ayn2op: Deferred functions are LIFO, so the `(*MessagesText).endRegion` is called before `fmt.Fprintln`. It might be better to just use `Fprintln` instead of deferring it for readability.

ayn2op (approved):
Thank you! 🚀

PR #436, opened 2024-09-22
Store ID instead of index of message

### LINKED ISSUE #407: Wrong reply
When I hit r to reply after selecting, as the chat moves it replies the next message or to the next number of message defined by how many new messages have been send. Basically, reply isnt stable please look into this, thanks!.

### LINKED ISSUE #434: Change message pointing logic to use IDs
In relation to #407, and because regions already use them, message pointing logic should be changed from being index-based (via `selectedMessage`) to ID based. From there, all logic that points to a message (SelectNext/Prev, replies, etc) will get the proper one without the need to adjust selectedMessage manually (particularly on message creation and deletion).

Retropaint:
Closes #434 and #407 

`messagesText.selectedMessage` and `messageInput.replyMessageIdx` have been replaced with `messagesText.selectedMessageID` and `messagesInput.replyMessageID`, respectively. Of course, everything else pertaining to the old vars and logic has been changed as well. 

SelectPrev & Next now use a new func, `getSelectedMessageIdx`, to get index of the current message pointed by the ID to move around. 

Since replies now also use ID, there are no more misalignment issues. This closes #407 but also ensure that this kind of bug won't be a thing going forward.

Hopefully the revamp isn't too crazy. :P

### DIFF
--- a/cmd/message_input.go
+++ b/cmd/message_input.go
@@ -24,7 +24,6 @@ type MessageInput struct {
 func newMessageInput() *MessageInput {
 	mi := &MessageInput{
 		TextArea:        tview.NewTextArea(),
-		replyMessageID: 0,
 	}
 
 	mi.SetTextStyle(tcell.StyleDefault.Background(tcell.GetColor(cfg.Theme.BackgroundColor)))
--- a/cmd/message_input.go
+++ b/cmd/message_input.go
@@ -18,12 +18,13 @@ import (
 type MessageInput struct {
 	*tview.TextArea
 	replyMessageIdx int
+	replyMessageID discord.MessageID 
 }
 
 func newMessageInput() *MessageInput {
 	mi := &MessageInput{
 		TextArea:        tview.NewTextArea(),
-		replyMessageIdx: -1,
+		replyMessageID: 0,
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/state.go
```diff
@@ -89,7 +89,7 @@ func (s *State) onMessageCreate(m *gateway.MessageCreateEvent) {
 
 func (s *State) onMessageDelete(m *gateway.MessageDeleteEvent) {
 	if mainFlex.guildsTree.selectedChannelID == m.ChannelID {
-		mainFlex.messagesText.selectedMessage = -1
+		mainFlex.messagesText.selectedMessageID = 0
```
ayn2op, line 92: Use `discord.NullMessageID` instead.

cmd/message_input.go
```diff
@@ -18,12 +18,13 @@ import (
...
 	*tview.TextArea
 	replyMessageIdx int
+	replyMessageID discord.MessageID 
 }
 
 func newMessageInput() *MessageInput {
 	mi := &MessageInput{
 		TextArea:        tview.NewTextArea(),
-		replyMessageIdx: -1,
+		replyMessageID: 0,
```
ayn2op: Remove this line as Go already initializes unspecified fields of a `struct` as their zero value.

ayn2op (approved):
:rocket: LGTM, thank you for contributing!

PR #441, opened 2024-09-23
fix: highlight messages using mouse

b2ag:
Hi,

this should synchronize message selections done by mouse with the keyboard way of selecting messages. Thanks to the recent rework of the logic in the keyboard selection, it was quite easy to implement now.

It sets `messagesText.selectedMessageID` after it already was set in `messagesText._select`. So it's done twice for keyboard inputs. I couldn't be bothered to change something about it though as it doesn't affect functionality. Maybe it has some potential for weird race conditions. :shrug: 

Best,

### DIFF
--- a/cmd/messages_text.go
+++ b/cmd/messages_text.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"log/slog"
 	"slices"
+	"strconv"
 	"time"
 
 	"github.com/atotto/clipboard"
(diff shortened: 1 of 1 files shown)


cmd/messages_text.go
```diff
@@ -280,6 +284,17 @@ func (mt *MessagesText) _select(name string) {
 	mt.ScrollToHighlight()
 }
 
+func onHighlighted(added, removed, remaining []string) {
```
ayn2op: Please convert this into a receiver function:
```go
func (mt *MessagesText) onHighlighted(...) {}

b2ag: I was actually trying this at first but failed because I'm new to programming in go. Understood what was missing today after some sleep.

ayn2op (approved):
LGTM!

PR #443, opened 2024-09-26
Proper msg caching

Retropaint:
This PR replaces the default 100 cabinet message limit imposed by Arikawa, to the message limit specified in the user config. This prevents loading more messages than necessary since Cabinet retrieves from API if things are missing, wasting half of the messages being sent if using the default config (of 50 messages).

Somewhat unrelated but still applies: also added a limit of 100 to MessagesLimit

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -4,6 +4,7 @@ import (
 	"log/slog"
 
 	"github.com/rivo/tview"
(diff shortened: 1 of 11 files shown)


Retropaint:
Since I'm considering #444, this issue should be resolved along with it.

PR #451, opened 2024-10-06
Individual message rendering overhaul

### LINKED ISSUE #444: Handle messages individually
Currently, messages are rendered using a single tview.TextView. Since TextViews do not support editing its own text, any updates to messages can only be done by updating it all over again. This has 2 issues that, while not major now, will cause problems as the project grows in complexity:

- **Performance** - having to update all messages for any kind of edit (be it as small as opening spoilers), will cause more slowdown as support for loading more messages is added (#308)
- **Overcomplicated code** - any and all logic pertaining to editing messages has to pass through the creation process first. This means that if, for example, support for adding spoilers were added, it needs to be tracked separately so that the creation logic knows to render it as it should, rather than simply having to update a single message.

Proposal:
Create a new tview [Primitive](https://pkg.go.dev/github.com/rivo/tview#Primitive) for handling and rendering messages. Here's how it will fix the above issues:

- Messages will only be rendered as much as the screen can actually show, or a config limit. Any message updates will simply require a redraw for the actual message, rather than everything.
- Everything pertaining to messages will be handled on a per-message basis within the code, and will not require some weird way of tracking what and where things are done. Want to open a message's spoiler? Just make it redraw only that message with them open.

Retropaint:
Closes #444 

This is quite a hefty overhaul to how messages are rendered. Messages are now individual textViews rendered in a regular box, as opposed to one big textView. The benefits of this are described in the issue.

However, there is a concern I have for my implementation of this, which is the custom scrolling logic. Since tview boxes don't support scrolling, this has to be manually done. My implementation is *very* verbose but hopefully simple to understand and maintain. I had already tried a more 'compact' solution as seen in older commits but it was fragile in every sense of the word.

Note that this PR is strictly for the overhaul itself (made in the 'easiest' way possible) and doesn't have performance optimization. Attempting it was a time sink which I don't want to get into for now. I may add easy ones to implement into the PR if I find them.

Not sure if this is necessary to add, but I welcome any feedback on this system, particularly the scrolling logic (which I think can still be improved, I've just run out of ideas and time).


Retropaint:
Marked as draft as I hadn't yet fixed detecting text line counts for those with words longer than the box width.

Retropaint:
Scrolling logic has improved, not as verbose and still just as stable. Also some optimization with not calling Render on off-screen messages and caching line counts (will need re-calculating for edited messages but thats for another time).

Retropaint:
There are a few line count padding issues to fix with code blocks but it requires changing renderer.go, and I don't want to add more files changed to this PR lest it have more conflicts down the line.

@ayn2op Please let me know what this PR needs before it can be merged. If it will take a while to review that's fine, just as long as it's merged or closed at some point.

Retropaint:
noticed a bug where messages won't render on their own without any input, since messagesText is now a Box and doesn't get implicitly called to redraw. Will fix a bit later.

PR #527, opened 2025-04-14
feat: add toast notifications

### LINKED ISSUE #466: [Feature Request] Message-/ping-notifications
I would love to see a notification for new incoming direct-messages or/and server-pings, I haven't been able to receive notifications for now (using linux KDE plasma/steam deck in desktop mode) but would love to see a implementation of system-wide notifications if possible!

ludmila-lovelace:
create a toast notification (using https://github.com/gen2brain/beeep) when a message is received. supports custom configuration, guild nicknames, muted channels, and avatar caching

closes #466 

### Server message
image

### Private message (with attachment)
image

### DIFF
--- a/internal/config/notifications.go
+++ b/internal/config/notifications.go
@@ -1,0 +1,25 @@
+package config
+
+type (
+	Notifications struct {
+		Enabled   bool   `toml:"enabled"`
+		Duration  int    `toml:"duration"`
+		PlayChime Chimes `toml:"play_chime"`
+	}
+
+	Chimes struct {
+		Enabled    bool `toml:"enabled"`
+		OnlyOnPing bool `toml:"only_on_ping"`
+	}
+)
+
+func defaultNotifications() Notifications {
+	return Notifications{
+		Enabled:  true,
+		Duration: 500,
+		PlayChime: Chimes{
+			Enabled:    true,
+			OnlyOnPing: true,
+		},
+	}
+}
--- a/internal/config/notifications.go
+++ b/internal/config/notifications.go
@@ -2,12 +2,12 @@ package config
 
 type (
 	Notifications struct {
-		Enabled   bool   `toml:"enabled"`
-		Duration  int    `toml:"duration"`
-		PlayChime Chimes `toml:"play_chime"`
+		Enabled  bool   `toml:"enabled"`
+		Duration int    `toml:"duration"`
+		Sounds   Sounds `toml:"sounds"`
 	}
 
-	Chimes struct {
+	Sounds struct {
 		Enabled    bool `toml:"enabled"`
 		OnlyOnPing bool `toml:"only_on_ping"`
 	}
@@ -17,7 +17,7 @@ func defaultNotifications() Notifications {
 	return Notifications{
 		Enabled:  true,
 		Duration: 500,
-		PlayChime: Chimes{
+		Sounds: Sounds{
 			Enabled:    true,
 			OnlyOnPing: true,
 		},
--- a/internal/config/notifications.go
+++ b/internal/config/notifications.go
@@ -2,12 +2,12 @@ package config
 
 type (
 	Notifications struct {
-		Enabled  bool   `toml:"enabled"`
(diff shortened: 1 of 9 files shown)


ayn2op:
I just ran this locally on macOS; it does not display the notification banner/toast, but it does play the macOS beep sound.

ludmila-lovelace:
@ayn2op might be related to this - https://github.com/gen2brain/beeep/issues/67? let me know if the fix mentioned there works for macOS

i will try to test on Windows later today as well

edit: beep sound and banner are both working on Windows 11
image

ludmila-lovelace:
update: i set up a macOS VM to test this and, while the workaround fixes the issue, it's a bit tedious and the notification isn't very pretty. instead, i'm thinking we can use [gosx-notifier](https://github.com/deckarep/gosx-notifier) for Darwin builds (the library is unmaintained but i wasn't able to find a better alternative)

on first notification:
Screenshot_20250416_120517

after allowing:
Screenshot_20250416_115948

internal/config/notifications.go
```diff
@@ -0,0 +1,25 @@
+package config
+
+type (
+	Notifications struct {
+		Enabled   bool   `toml:"enabled"`
+		Duration  int    `toml:"duration"`
+		PlayChime Chimes `toml:"play_chime"`
```
ayn2op: What do you think about renaming this field to `sounds` (play_chime feels too specific).

internal/notifications/desktop_toast_darwin.go
```diff
@@ -0,0 +1,19 @@
...
+	gosxnotifier "github.com/deckarep/gosx-notifier"
+)
+
+func sendDesktopNotification(title string, body string, image string, playSound bool, _ int) error {
+	notify := gosxnotifier.NewNotification(body)
+	notify.Title = title
+	notify.ContentImage = image
+
+	if playSound {
+		notify.Sound = gosxnotifier.Ping
```
ayn2op: Instead of `gosxnotifier.Ping`, `Default` might be far more applicable here.
Reference: <https://github.com/mirtchovski/gosxnotifier/blob/master/gosx-notifier.go#L12>

internal/notifications/notifications.go
```diff
@@ -0,0 +1,118 @@
...
+	resp, err := http.Get(url)
+	if err != nil {
+		return "", err
+	}
+	defer resp.Body.Close()
+
+	_, err = io.Copy(image, resp.Body)
+	if err != nil {
+		return "", err
+	}
```
ayn2op: You can inline this if statement:
```suggestion
	if _, err := io.Copy(image, resp.Body); err != nil {
		return "", err
	}
```

internal/notifications/notifications.go
```diff
@@ -0,0 +1,118 @@
...
+
+	path = filepath.Join(path, consts.Name)
+	if err := os.MkdirAll(path, os.ModePerm); err != nil {
+		return "", err
+	}
+
+	path = filepath.Join(path, avatarHash+".png")
```
ayn2op, line 95:
```suggestion
	path = filepath.Join(path, avatarHash, ".png")
```

ludmila-lovelace: i don't think `filepath.Join` treats extensions differently to regular path arguments (see [example](https://go.dev/play/p/1g7LCprw8pF)) so i think this one should be left as-is

internal/config/notifications.go
```diff
@@ -0,0 +1,25 @@
+package config
+
+type (
+	Notifications struct {
+		Enabled  bool   `toml:"enabled"`
+		Duration int    `toml:"duration"`
+		Sounds   Sounds `toml:"sounds"`
```
ayn2op: Now that I think about it, it might be better to rename this field to `sound` (representing singular noun instead).

ayn2op: Everything else looks good, this PR can be merged after this change is addressed.

ayn2op (approved):
LGTM! 🚀 🥳

PR #542, opened 2025-05-14
feat(config): generate default config on first launch

klrmngr:
Generates the `~/.config/discordo/config.toml` file on startup if it does not exist


ayn2op:
This is intentional, Discordo is currently in alpha and while I try to minimize breaking changes, the configuration fields are changing without notice.

PR #544, opened 2025-05-21
feat: add ability to disable markdown

ted-mundy:
Added the ability to disable markdown rendering in the config

Before: _we_
image

After: `*we*`
image


ayn2op (approved):
LGTM

PR #322, opened 2023-08-16
Add Nix flake

emanueljg:
This PR adds support for the [Nix](https://nixos.org/) build system.

PR #545, opened 2025-05-24
Add Nix

emanueljg:
Add support for the Nix build system. 

Supersedes #322.

Included are flake, package, home-manager module, and devshell. 

It makes sense to have an upstream-specific package definition because `src` will be different and is more likely to change, but the home-manager module is a temporary addition until it's downstreamed in the home-manager repo; there's no real reason to keep a home-manager module upstream, as it's a very standard generic `.settings` module.


emanueljg:
as of https://github.com/ayn2op/discordo/pull/545/commits/7dfc7cd76f32006ed25f74fc5a84ae691e05ae70, we go from a full `buildGoModule` drv to an override of downstream `pkgs.discordo`:

```nix
# ./nix/package.nix (pkgs.callPackage ./nix/package.nix { })
{ discordo
, lib
}: discordo.overrideAttrs {
  version = "git";

  src = let fs = lib.fileset; in fs.toSource {
    root = ../.;
    fileset = fs.unions [
      ../go.mod
      ../go.sum
      ../main.go
      ../cmd
      ../internal
    ];
  };

  vendorHash = "sha256-gEwTpt/NPN1+YpTBmW8F34UotowrOcA0mfFgBdVFiTA=";
}
```
@linuxmobile As the nixpkgs maintainer, does this construct seem reasonable to you? This way we don't waste time duplicating drv attrs, yet we're able to do local development using the local source. In this weird way, we as upstream are dependent on downstream, but as far as I know (let me know if I'm wrong!), there's no way for downstream `pkgs.discordo` to consume our derivation, so if we want deduplication I'm not sure there's an other way to do this.

PR #550, opened 2025-05-25
feat(config): add light border preset

xqrs:
The Arc version of these characters did not exists in the IBM-437 encoding,
which is used by the default Linux tty font.


ayn2op (approved):
Thanks! <3

PR #551, opened 2025-05-25
feat(config): include a default config file

xqrs:
This clears the internal structures:
defaultConfig, defaultKeys, and defaultTheme.

Then includes a config.toml and then parses it, instead of duplicating
data or letting a possible future config file and actual in-program
default values not being equal.

This also forces any future configs to have proper documentation
and an example attached to them that non-developers can read.

Fixes #459

PR #554, opened 2025-05-30
mentions

xqrs:
- **add nix**
- **add direnv folder to .gitignore**
- **deduplicate pkg with downstream overrideAttrs**
- **build(nix): update vendorHash**
- **feat(config): include a default config file (#551)**
- **fix(config): remove space after go:embed directive**
- **fix(config): import embed package**
- **refactor: gofmt**
- **Add mention support**


xqrs:
this sucks, I want to do REAL tui with proper tab completion and recommendations, not just colored text. I gonna work on an new version.

PR #555, opened 2025-06-01
feat: add mention support

### LINKED ISSUE #129: Feature request: List and tag users in a channel

xqrs:
Unlike the previous patch, this patch actually opens a page list with
possible candidates, it also includes the following features:

* Caching (no need to flood the gateway with redundent searches)
* Automatic suggestions (no need to press tab after typing `@`)
* Real tab completion (pressing tab will actually complete the username)
* Pressing Enter while completing will mention the selected username
* Automatic tab processing not blocking the main gorountine.

https://asciinema.org/a/iJzwtCss0SvSLqwgJDg1K19VN

Note: the current code is a mess because of the restrictive Tview API.

GetCursor returns *the displayed cursor position*, not the actual position
within the text and there's no direct way to get the actual characters
under the cursor without doing weird math with word wrapping.

If we want to simplify the code, Tview would have to provide GetCursorPos to
get the actual cursor position without funny word wrapping and
GetWordUnderCursor to get the word currently under the cursor.
If anybody here already knows and worked with Tview then please implement these
functions, it would make the code way faster and simpler.

I also thought of changing the bottom border character corners of the
candidates list to BoxDrawingsLightUpAndHorizontal (aka the ┴ character)
but Tview only supports one border for all primitives.

Closes #129, #483


internal/config/config.toml
```diff
@@ -119,10 +120,17 @@ channel_color = "white"
...
 reply_indicator = ">"
 forwarded_indicator = "<"
+
 author_color = "aqua"
 content_color = "white"
 emoji_color = "green"
 link_color = "blue"
 attachment_color = "yellow"
+
+[theme.candidates]
```
ayn2op: Perhaps, a toggle to turn on/off auto-show behavior of the candidates list? Neovim has that.

ayn2op:
Hey, @xqrs. Thanks for the contribution! You may wanna join the Discord server, there are tons of implementation details in there: https://discord.com/invite/VzF9UFn2aB

internal/cache/cache.go
```diff
@@ -0,0 +1,83 @@
+// This is not a real cache system, it only remembers what is the time
```
ayn2op: Maybe we can use this instead: https://github.com/hashicorp/golang-lru

xqrs:
I want to reimplement the cache system of this and also fix the slowness when having a "lot" of members (15 on my machine) and I also want to do fuzzy finding this time with displaynames, not just usernames

ayn2op:
Closes #129

cmd/message_input.go
```diff
@@ -105,6 +171,253 @@ func (mi *MessageInput) send() {
...
+				res.WriteByte(':')
+			}
+		}
+		return ast.WalkContinue, nil
+	})
+	return res.String()
+}
+
+func expandMentions(src []byte) []byte {
+	rx := regexp.MustCompile("@[a-zA-Z0-9._]+")
```
ayn2op: Move the `regexp.MustCompile` call outside of `expandMentions`, so it is not recompiled on every call.

```go
var mentionRegex = regexp.MustCompile(`@[a-zA-Z0-9._]+`)
```

cmd/message_input.go
```diff
@@ -105,6 +171,253 @@ func (mi *MessageInput) send() {
...
+	return rx.ReplaceAllFunc(src, func(in []byte) (out []byte) {
```
ayn2op: You may wanna look into `ReplaceAllStringFunc`:
```go
func (re *Regexp) ReplaceAllStringFunc(src string, repl func(string) string) string
```

xqrs:
i forced-pushed because the commit log was a mess made by a mistake

cmd/message_input.go
```diff
@@ -5,32 +5,48 @@ import (
...
+	"github.com/diamondburned/arikawa/v3/state"
 	"github.com/diamondburned/arikawa/v3/utils/json/option"
+	"github.com/diamondburned/ningen/v3/discordmd"
 	"github.com/gdamore/tcell/v2"
+	"github.com/yuin/goldmark/ast"
 	"github.com/rivo/tview"
 )
 
 const tmpFilePattern = consts.Name + "_*.md"
+var mentionRegex *regexp.Regexp = regexp.MustCompile("@[a-zA-Z0-9._]+")
```
ayn2op: Specifying the type is optional.
```suggestion
var mentionRegex = regexp.MustCompile("@[a-zA-Z0-9._]+")
```

PR #595, opened 2025-08-18
Switch from zenity to a tui file picker

voidpixe1:
output


ayn2op:
Instead of implementing it ourself, we could probably delegate it to an existing fuzzy-finder (fzf) or a terminal file browser (nnn).

ayn2op:
For reference, check out how external child process (editor) for editing text is executed (`(*messageInput).editor`).

voidpixe1:
> Instead of implementing it ourself, we could probably delegate it to an existing fuzzy-finder (fzf) or a terminal file browser (nnn).

i see but wouldn't that create more dependencies, i mean we are already importing [a fuzzy matcher for mentions](https://github.com/sahilm/fuzzy), can't we just add a file picker with search functionality within the tui?

voidpixe1:
although i can see that using an external program would be more flexible, cuz then customization would be easier

ossenthusiast:
Update?

voidpixe1:
> Update?

i am trying to add a search functionality to this, but my skill issue is getting in the way (not to mention that i basically copy pasted the entire functionality of showing attachment lists which was not a good idea, so i guess i will have to redo it from scratch)

voidpixe1:
my skill issue is getting in the way a lot, let me open another PR

PR #597, opened 2025-08-20
README: help with the first steps

starius:
- Provide a way to get a token.
 - Explain how to find chats after logging in.

This guidance unblocked me.


aricshow:
The server folder tip helped me, thank you. I found obtaining the token from localStorage (key `token`) in the browser to be easier than inspecting requests headers.

PR #605, opened 2025-09-26
feat: add confirmation modal for delete action (#575)

### LINKED ISSUE #575: delete confirm
When executing d, can it be confirmed again?

Is it possible to add a setting in [config.toml](https://github.com/ayn2op/discordo/blob/8886a1baad5aee77ce7efe501b4b416437e19c0b/internal/config/config.toml#L78), for example:

```toml
delete = "Rune[d],Confirm('delete this message?')"
```

If Confirm is added afterward, it requires confirmation before the action is performed.

JACTyt:
Hi, I created a PR that solves issue #575 .
Added delete_confirmation variable (true/false). 
If true, then there will be confirmation popup in the center of the screen with buttons Yes or No. 
If false, then no confirmation popup will appear. Simple deletion without confirmation (usage depends on user preferences).
Have some doubts about func  showModal() placement.

Closes #575


cmd/application.go
```diff
@@ -15,6 +16,7 @@ const (
 	flexPageName            = "flex"
 	mentionsListPageName    = "mentionsList"
 	attachmentsListPageName = "attachmentsList"
+	modalPageName           = "Modal"
```
ayn2op: Rename the const name to something more descriptive: `confirmModalPageName`:

```go
const confirmModalPageName = "confirmModal"
```

internal/config/config.toml
```diff
@@ -70,6 +70,7 @@ reply = "Rune[r]"
 reply_mention = "Rune[R]"
 cancel = "Esc"
 delete = "Rune[d]"
+delete_confirmation = true
```
ayn2op: Perhaps, `Rune[d]` can be used for deleting messages with confirmation and `Rune[D]` for deleting them without. This is similar to what `reply` and `reply_mention`. What do you think?

internal/config/keys.go
```diff
@@ -40,9 +40,10 @@ type (
 		Reply        string `toml:"reply"`
 		ReplyMention string `toml:"reply_mention"`
 
-		Cancel string `toml:"cancel"`
-		Delete string `toml:"delete"`
-		Open   string `toml:"open"`
+		Cancel             string `toml:"cancel"`
+		Delete             string `toml:"delete"`
+		DeleteConfirmation bool   `toml:"delete_confirmation"`
```
ayn2op: https://github.com/ayn2op/discordo/pull/605/files#r2383657890

cmd/application.go
```diff
@@ -154,14 +154,14 @@ func (a *application) onFlexInputCapture(event *tcell.EventKey) *tcell.EventKey
 	return event
 }
 
-func (a *application) showModal(prompt string, buttons []string, onDone func(label string)) {
+func (a *application) showConfirmModal(prompt string, buttons []string, onDone func(label string)) {
```
ayn2op, line 157: The `buttons` parameter can be removed and hardcoded to `[]string{"Cancel", "Confirm"}` instead.

JACTyt: In my opinion, hardcoding this part is not a good idea. We can reuse confirmModal for example to provide some debug Information or when opening web pages (confirm if you trust the link and want to open it, etc.). I will do as you wish. Please verify, do i need to hardcode it?

ayn2op (approved):
LGTM! Thank you <3

PR #606, opened 2025-09-27
Added "man page" & script to install man page

Gamedirection:
A manual page for quick reference to keybindings and configuration because I kept having to look at configs for the keybindings because I have the memory of a goldfish.


ayn2op:
AFAIK, manpages are Unix-only; Discordo, on the other hand, supports all modern operating systems. Please consider opening an issue before a PR.

PR #621, opened 2025-11-15
feat: show error modal on errors

xqrs:


cmd/application.go
```diff
@@ -2,20 +2,23 @@ package cmd
...
 	"github.com/gdamore/tcell/v2"
+	"golang.design/x/clipboard"
 )
 
 const (
 	flexPageName            = "flex"
 	mentionsListPageName    = "mentionsList"
 	attachmentsListPageName = "attachmentsList"
-	confirmModalPageName    = "confirmModal"
+	ModalPageName           = "Modal"
```
ossenthusiast: modalPageName

cmd/application.go
```diff
@@ -202,22 +205,60 @@ func (a *application) focusNext() {
...
+			res.WriteString(msg)
+			res.WriteString("\nReason: ")
+			res.WriteString(err)
+			for i := 0; i < len(info)-1; i += 2 {
+				res.WriteRune(rune('\n'))
+				res.WriteString(info[i].(string))
+				res.WriteString(": ")
+				res.WriteString(fmt.Sprintf("%#v", info[i+1]))
+			}
+			go clipboard.Write(clipboard.FmtText, res.Bytes())
```
ossenthusiast: Just write `modal.GetText()` in clipboard instead of writing to an intermediate buffer. The callback can forward the modal: `func(modal *tview.Modal, label string)`

cmd/application.go
```diff
@@ -202,22 +205,60 @@ func (a *application) focusNext() {
...
+func (a *application) showErrorModal(msg, err string, info ...any) {
+	a.showModal("[ ERROR ]", msg + "\nReason: " + err, []string{"Copy", "OK"}, func(label string) {
+		if label == "Copy" {
+			res := bytes.Buffer{}
+			res.WriteString(msg)
+			res.WriteString("\nReason: ")
+			res.WriteString(err)
+			for i := 0; i < len(info)-1; i += 2 {
+				res.WriteRune(rune('\n'))
+				res.WriteString(info[i].(string))
```
ossenthusiast: This can panic? Since values of log statement are not guaranteed to be `string`, but keys are.

```go
panic: interface conversion: interface {} is int, not string
```

ossenthusiast: Nvm; correction: info[i] refers to be the key itself, not the value.

cmd/application.go
xqrs: Not everything within the modal will be copied to the clipboard and the other way around is also true.

The modal only contains a simple error message, the full info can only be accessed through the Copy feature.

cmd/application.go
```diff
@@ -202,22 +205,60 @@ func (a *application) focusNext() {
...
+}
+
+func (a *application) showErrorModal(msg, err string, info ...any) {
+	a.showModal("[ ERROR ]", msg + "\nReason: " + err, []string{"Copy", "OK"}, func(label string) {
+		if label == "Copy" {
+			res := bytes.Buffer{}
+			res.WriteString(msg)
+			res.WriteString("\nReason: ")
+			res.WriteString(err)
+			for i := 0; i < len(info)-1; i += 2 {
```
ossenthusiast: This skips the last element silently if the slice is odd-length, but doesn’t prevent someone from passing malformed key/value pairs

cmd/application.go
```diff
@@ -202,22 +205,60 @@ func (a *application) focusNext() {
...
+				res.WriteRune(rune('\n'))
```
ossenthusiast: res.WriteRune('\n')

xqrs: ahh yes.

cmd/application.go
xqrs: Only if you passed non-string as the title for one of the info elements (which is never done).

cmd/messages_list.go
```diff
@@ -564,18 +563,18 @@ func (ml *messagesList) reply(mention bool) {
...
 	me, err := discordState.Cabinet.Me()
 	if err != nil {
-		slog.Error("failed to get client user (me)", "err", err)
+		app.onError("Failed to get client user (me)", err)
 		return
 	}
 
 	if message.Author.ID != me.ID {
-		slog.Error("failed to edit message; not the author", "channel_id", message.ChannelID, "message_id", message.ID)
+		app.onError("Failed to edit message", errors.New("You are not the author"), "channel_id", message.ChannelID, "message_id", message.ID)
```
ossenthusiast, line 577: `err` cannot simply be "You are not the author." It feels awkward:
```log
time=2025-11-21T16:38:14.775-05:00 level=ERROR msg="Failed to edit message" err="You are not the author" ...
```
Imo, err should not be provided and that case should be handled by `onError`.

ossenthusiast: Perhaps, pass `"err", err` and remove the error parameter from onError func.

xqrs: I prefer a reason for errors. "Cannot edit msg" isn't much useful without "Reason: You are not the author."

ossenthusiast: Yes, but both can be merged and err can be nil then. In this case, `err` is just `You are not the reason`, which feels incomplete.

ossenthusiast: msg = failed to edit message; you are not the author
err = nil

Since, in slog.Info call only msg is a non-nilable argument (string).

PR #642, opened 2025-12-01
Improved reply indicators

qasimstatic:
## Summary
- Replaced '>' with '󱞩' reply indicator(nf-md-arrow_right_bottom)
- Move indicator from referenced message to actual reply
- Properly dim referenced messages for better visual hierarchy
- Update forwarded indicator(nf-md-arrow_right_top) to matching '󱞫' icon

## Changes
- Modified message rendering logic in `cmd/messages_list.go`
- Updated default indicators in `internal/config/config.toml`
- Improved visual distinction between referenced messages and actual replies

## Testing
- Verified reply indicators appear before actual reply messages
- Confirmed referenced messages are properly dimmed
- Tested with various message types (replies, regular messages, forwarded)

## Summary by CodeRabbit

* **New Features**
  * Enhanced message reply display with improved formatting and positioning of reply indicators.

* **Style**
  * Updated reply and forwarded message indicators to Unicode arrows (↳ and ↱) for clearer visual distinction.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


xqrs:
image
The character does not render for me, and there's no search result for it.

qasimstatic:
would you mind checking 
nf-md-arrow_right_bottom
nf-md-arrow_right_top
[here](https://www.nerdfonts.com/cheat-sheetl)

xqrs:
It's a nerd-font specific glyph, huh. I don't think @ayn2op would  like that.

qasimstatic:
@xqrs Yeah thank you for telling me that I also assumed that would be the case so we could use the usual unicode '↱' and '↳'

ossenthusiast:
image

This looks rather odd, imo. Contrary to the official client, the indicator is now after the timestamp, so it is not immediately obvious to the user that the message is a reply.

Screenshot 2025-12-01 at 2 17 06 PM

qasimstatic:
Oh yeah I am trying to get the hierarchy in which discord displays their messages.
If one person stacks 3 messages under his message his name doesn't repeat.
Mb I was testing with time off because of this.
I'll make some edits for the time and reply indicator to check if it looks good

PR #649, opened 2025-12-07
fix(message_input): Make mention list work in DMs

### LINKED ISSUE #637: Can't ping people on dm's
When i do the @ and put a single letter everithing crashes

xqrs:
Before this, typeing `@` in DMs hanged the entire client.
This patch fixes that and implements proper handling of DMs (DM and GroupDMs)

Edit:
Fixes #637


cmd/guilds_tree.go
```diff
@@ -142,8 +142,6 @@ PARENT_CHANNELS:
 }
 
 func (gt *guildsTree) onSelected(node *tview.TreeNode) {
-	app.chatView.messageInput.reset()
```
ossenthusiast, line 145: Is this supposed to not happen at all? I think the message input should be reset when a new channel is selected. So ig move this to in `case discord.ChannelID` branch?

cmd/guilds_tree.go
```diff
@@ -183,6 +181,9 @@ func (gt *guildsTree) onSelected(node *tview.TreeNode) {
 			app.chatView.messagesList.requestGuildMembers(guildID, messages)
 		}
 
+		app.chatView.selectedChannel = channel
+		app.chatView.selectedGuildID = channel.GuildID
```
ossenthusiast: We can probably get rid of selectedGuildID since `selectedChannel.GuildID` does the exact same thing.

cmd/guilds_tree.go
xqrs: `messageInput.reset()` (clear input box) should be called only when switching channels.

The user may actually be browsing other servers while chatting with someone.

cmd/guilds_tree.go
```diff
@@ -197,9 +197,7 @@ func (gt *guildsTree) onSelected(node *tview.TreeNode) {
 			app.SetFocus(app.chatView.messageInput)
 		}
 
-		app.chatView.selectedChannelID = channel.ID
-		app.chatView.selectedGuildID = channel.GuildID
-	case nil: // Direct messages
+	case nil: // Direct messages folder
```
ayn2op, line 200: unrelated change

ayn2op: This does not make any sense, "Direct Messages" node is a `*tview.TreeNode`, not a folder.

xqrs: When I was reading the code some late time of the night, I thought `case nil: // Direct messages` meant *DMs* as in it's selected when you choose one of the DMs channel. What should we actually call it? "Direct messages" is ambiguous between a list of channels and just one channel.

PR #651, opened 2025-12-08
feat(message_input): separate key for cancelling reply and deleting i…

nyxmeowmeow:
i think its a little bit confusing and annoying that these are using the same key, especially since theres no undo button. sorry if im doing something wrong, im still new to this stufft.

### DIFF
--- a/internal/config/config.toml
+++ b/internal/config/config.toml
@@ -89,8 +89,10 @@ yank_id = "Rune[i]"
 # paste from clipboard (supports both text and images)
 paste = "Ctrl+V"
 send = "Enter"
-# Remove existing text or cancel reply.
+# Cancel reply.
 cancel = "Esc"
+# Remove existing text.
+reset = "Ctrl+X"
 # Complete usernames when mentioning
 tab_complete = "Tab"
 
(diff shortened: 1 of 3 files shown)


internal/config/config.toml
```diff
@@ -89,8 +89,10 @@ yank_id = "Rune[i]"
 # paste from clipboard (supports both text and images)
 paste = "Ctrl+V"
 send = "Enter"
-# Remove existing text or cancel reply.
+# Cancel reply.
 cancel = "Esc"
+# Remove existing text.
+reset = "Ctrl+X"
```
ossenthusiast: - Ctrl+X is already used for `cut`: https://github.com/ayn2op/tview/blob/07345f1515187e69c35a7c5cc95558bda71a0e60/textarea.go#L2252
- `TextArea` widget already supports deleting the current line using `Ctrl+U`: https://github.com/ayn2op/tview/blob/master/textarea.go#L2237

nyxmeowmeow:
previous commit broke the cancel functionality. now the cancel button just cancels the reply. also removed the redundant reset key

PR #657, opened 2025-12-09
feat(clipboard): add wayland support

xqrs:
Introducing: `internal/clipboard`, a `golang.design/x/clipboard`-compatible
clipboard wrapper with Wayland support using `wl-copy` and `wl-paste`.

### DIFF
--- a/internal/clipboard/clipboard_linux.go
+++ b/internal/clipboard/clipboard_linux.go
@@ -11,26 +11,28 @@ import (
 )
 
 var wayland bool
+var inited bool
 
 func Init() error {
-	if wayland = checkWaylandEnv(); !wayland {
+	if _, ok := os.LookupEnv("WAYLAND_DISPLAY"); !ok {
+		inited = true
 		return designClipb.Init()
 	}
+	if _, err := exec.LookPath("wl-copy"); err != nil {
+		return err
+	}
+	if _, err := exec.LookPath("wl-paste"); err != nil {
+		return err
+	}
+	wayland = true
+	inited = true
 	return nil
 }
 
-func checkWaylandEnv() bool {
-	if _, ok := os.LookupEnv("WAYLAND_DISPLAY"); !ok {
-		return false
-	}
-	if _, err := exec.LookPath("wl-copy"); err != nil {
-		return false
-	}
-	_, err := exec.LookPath("wl-paste")
-	return err == nil
-}
-
 func Read(t Format) []byte {
+	if !inited {
+		return nil
+	}
 	if !wayland {
 		return designClipb.Read(designClipb.Format(t))
 	}
@@ -47,6 +49,9 @@ func Read(t Format) []byte {
 }
 
 func Write(t Format, buf []byte) <-chan struct{} {
+	if !inited {
+		return nil
+	}
 	if !wayland {
 		return designClipb.Write(designClipb.Format(t), buf)
 	}
--- a/internal/clipboard/clipboard_linux.go
+++ b/internal/clipboard/clipboard_linux.go
@@ -34,14 +34,9 @@ func Read(t Format) []byte {
 	if !wayland {
 		return designClipb.Read(designClipb.Format(t))
 	}
-	var typ string
-	switch t {
-	case FmtText:
-		typ = "text/plain;charset=utf-8"
-	case FmtImage:
-		typ = "image"
-	}
-	cmd := exec.Command("wl-paste", "--no-newline", "--type", typ)
+	// -n: Don't print a newline at the end
+	// -t type: MIME type specifier
+	cmd := exec.Command("wl-paste", "-nt", formatToType(t))
 	outBuffer := bytes.Buffer{}
 	cmd.Stdout = &outBuffer
 	if err := cmd.Run(); err != nil {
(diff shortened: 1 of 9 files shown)


xqrs:
I did some testing with Weston, it worked. More testing should be done with different setups to see how the code reacts to different states.

internal/clipboard/clipboard_linux.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+func checkWaylandEnv() bool {
+	if typ := os.Getenv("XDG_SESSION_TYPE"); typ != "wayland" {
+		if _, ok := os.LookupEnv("WAYLAND_DISPLAY"); !ok {
+			return false
+		}
+	}
+	_, err := exec.LookPath("wl-copy")
+	return err == nil
```
ayn2op: This is not necessary: `wl-clipboard` (`wl-copy` and `wl-paste`) expects a `WAYLAND_DISPLAY` anyway. Just checking for `WAYLAND_DISPLAY` should be enough.

- https://man.archlinux.org/man/extra/wl-clipboard/wl-clipboard.1.en#WAYLAND_DISPLAY
- https://man.archlinux.org/man/extra/wayland-docs/wl_display.3.en#int_wl_display_add_socket_(struct_%3Cb%3Ewl_display%3C/b%3E_*_display,_const_char_*_name)
- https://github.com/atotto/clipboard/blob/1438b322bde482cc6decd8bc53dbe2eece57668b/clipboard_unix.go#L52

xqrs: ok

internal/clipboard/clipboard_linux.go
```diff
@@ -0,0 +1,70 @@
...
+	return nil
+}
+
+func checkWaylandEnv() bool {
+	if typ := os.Getenv("XDG_SESSION_TYPE"); typ != "wayland" {
+		if _, ok := os.LookupEnv("WAYLAND_DISPLAY"); !ok {
+			return false
+		}
+	}
+	_, err := exec.LookPath("wl-copy")
```
ayn2op: Check for both `wl-copy` and `wl-paste` using `exec.LookPath`: https://github.com/atotto/clipboard/blob/1438b322bde482cc6decd8bc53dbe2eece57668b/clipboard_unix.go#L56

EDIT: It would not make sense for only one of them to be present though.

xqrs: *both?* ok

internal/clipboard/clipboard_linux.go
```diff
@@ -0,0 +1,71 @@
...
+	if !wayland {
+		return designClipb.Read(designClipb.Format(t))
+	}
+	var typ string
+	switch t {
+	case FmtText:
+		typ = "text/plain;charset=utf-8"
+	case FmtImage:
+		typ = "image"
+	}
```
ayn2op: since this is duplicate in both `Read` and `Write`, a `formatToType` function might be better here

xqrs: ok boss

internal/clipboard/clipboard_linux.go
```diff
@@ -0,0 +1,71 @@
...
+func checkWaylandEnv() bool {
+	if _, ok := os.LookupEnv("WAYLAND_DISPLAY"); !ok {
+		return false
+	}
+	if _, err := exec.LookPath("wl-copy"); err != nil {
+		return false
+	}
+	_, err := exec.LookPath("wl-paste")
+	return err == nil
+}
```
ayn2op: LookPath returns `(string, error)`, return the error from `Init`. It might be better to inline `checkWaylandEnv`:
```go
func Init() error {
    if _, ok := os.LookupEnv("WAYLAND_DISPLAY"); !ok {
        return designClipb.Init()
    }
    
    if _, err := exec.LookPath("wl-copy"); err != nil {
        return err
    }
    if _, err := exec.LookPath("wl-paste"); err != nil {
        return err
    }
	return nil
}
```

xqrs: oh, nice. never thought of that

internal/clipboard/clipboard_linux.go
```diff
@@ -0,0 +1,76 @@
...
+	if !wayland {
+		return designClipb.Read(designClipb.Format(t))
+	}
+	// -n: Don't print a newline at the end
+	// -t type: MIME type specifier
+	cmd := exec.Command("wl-paste", "-nt", formatToType(t))
+	outBuffer := bytes.Buffer{}
+	cmd.Stdout = &outBuffer
+	if err := cmd.Run(); err != nil {
+		slog.Error("failed to write to clipboard", "err", err)
```
ayn2op: write -> read

ayn2op:
```suggestion
		slog.Error("failed to read clipboard", "err", err)
```

internal/clipboard/clipboard_linux.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"bytes"
+	designClipb "golang.design/x/clipboard"
+	"log/slog"
+	"os"
+	"os/exec"
+)
+
+var wayland bool
+var inited bool
```
ayn2op: I don't think we need `inited`.

ossenthusiast:
```patch
diff --git a/README.md b/README.md
index ae03f5b..821eb1f 100644
--- a/README.md
+++ b/README.md
@@ -44,10 +44,7 @@ go build .
 
 ### Wayland clipboard support
 
-`x11-dev` is required for X11 clipboard compatibility:
-
-- Ubuntu: `apt install xwayland`
-- Arch Linux: `pacman -S xorg-xwayland`
+`wl-clipboard` is required for clipboard support.
 
 ## Usage
```
 

I think you should add this info to readme, @ayn2op WDYT?

ayn2op (approved):
Thank you! <3

PR #658, opened 2025-12-09
fix(chatview): conditionally focus message input for focus_previous and focus_next keys

ayn2op:


xqrs:
What does this do? I don't notice a difference between the behavior of this branch and `main`

ayn2op:
> What does this do? I don't notice a difference between the behavior of this branch and `main`

It no longer focuses on the message input if it is disabled, this was not the case before.

PR #663, opened 2025-12-12
feat: add toggle for colored nicknames

n-1-c-k-y:
Adds toggle for colored nicknames for people that prefer the nicknames just white


ayn2op:
`colored_member_names` (name = username/global name/nickname) sounds much more accurate, since a member may or may not have a nickname; color is independent of nickname of a member. WDYT?

n-1-c-k-y:
> `colored_member_names` (name = username/global name/nickname) sounds much more accurate, since a member may or may not have a nickname; color is independent of nickname of a member. WDYT?

Good point, I've updated it.

PR #665, opened 2025-12-16
feat: select unread messages

### LINKED ISSUE #664: Keyboard shortcut to select unread guild/channel/direct message
It would be great to have some kind of keyboard shortcut to be able to select the next channel/guild/direct message with an unread message in there.

mstruebing:
selects the first unread channel and expands the guild if necessary with shortcut `u` (aka unread) while selected the guild tree.

Closes #664


cmd/guilds_tree.go
```diff
@@ -222,6 +222,70 @@ func (gt *guildsTree) onSelected(node *tview.TreeNode) {
 	}
 }
 
+// selectFirstUnread finds and selects the first unread channel node in the tree.
+func (gt *guildsTree) selectFirstUnread() {
+	root := gt.GetRoot()
+
+	var guildNode *tview.TreeNode
+	foundGuild := false
```
ossenthusiast, line 261: Redundant: you can just check if guildNode != nil for foundGuild.

cmd/guilds_tree.go
```diff
@@ -222,6 +222,70 @@ func (gt *guildsTree) onSelected(node *tview.TreeNode) {
...
+	root.Walk(func(node, parent *tview.TreeNode) bool {
+		// it continues walking without this
+		if foundGuild {
+			return false
+		}
```
ossenthusiast, line 266: Redundant: since you are already returning `false` when a guild a found, so the Walk func returns early anyway.

cmd/guilds_tree.go
```diff
@@ -222,6 +222,70 @@ func (gt *guildsTree) onSelected(node *tview.TreeNode) {
...
+		if guildID, ok := node.GetReference().(discord.GuildID); ok {
+			indication := discordState.GuildIsUnread(guildID, ningen.GuildUnreadOpts{UnreadOpts: ningen.UnreadOpts{IncludeMutedCategories: true}})
+			if indication == ningen.ChannelMentioned || indication == ningen.ChannelUnread {
+				guildNode = node
+				foundGuild = true
+				return false
+			}
+		}
```
ossenthusiast, line 274: This only handles guilds, not private channels (DMs), check `state.go:80` for example:
```go
			switch node.GetReference() {
			case event.GuildID:
				node.SetTextStyle(app.chatView.guildsTree.getGuildNodeStyle(event.GuildID))
				guildNode = node
				return false
			case event.ChannelID:
				// private channel
				if !event.GuildID.IsValid() {
					style := app.chatView.guildsTree.getChannelNodeStyle(event.ChannelID)
					node.SetTextStyle(style)
					return false
				}
			}
```

PR #671, opened 2025-12-21
feat(config): add toggleable auto focus on channel select

### LINKED ISSUE #667: Feature request: Toggleable auto focus in case of panes
I noticed when I enter a channel or something it auto focuses on the chat pane for example. But what if the user only wants to preview the channel and read it? I think it would be useful to implement something like a boolean true or false in the TOML to do something like this on configuration end.

rajanarahul93:
# Title
feat(config): add toggleable auto focus on channel select

## Summary
This PR implements a new configuration option `auto_focus` (defaulting to `true`) that allows users to disable the automatic focus switch to the message input when selecting a channel.

## Root cause
Previously, selecting a channel in the Guilds Tree unconditionally moved focus to the Message Input (inside `cmd/guilds_tree.go`), preventing users from previewing channels without losing their place in the tree.

## Changes
- **internal/config**: Added `AutoFocus` field to Config struct and `config.toml`.
- **cmd/guilds_tree.go**: Wrapped the `app.SetFocus` call in `onSelected` with a check for `gt.cfg.AutoFocus`.

## Testing
- Verified that `auto_focus = true` (default) preserves existing behavior (focus jumps to input).
- Verified that `auto_focus = false` keeps focus on the Guilds Tree after channel selection.
- Verified build passes with `go build .`

## Risk & Rollback
- **Risk:** Low. Default behavior is preserved via the embedded `config.toml`.
- **Rollback:** Revert this commit.

## Related
- Closes #667


ayn2op (approved):
LGTM! Thank you <3

PR #673, opened 2025-12-26
fix(mentionsList)!: Use item styles instead of style tags

xqrs:
requires https://github.com/ayn2op/tview/pull/3


xqrs:
before:
image
after:
image

xqrs:
No longer applicable after the change to the new TextSegments API

PR #676, opened 2025-12-29
fix(http/props): mask launch_signature to prevent client mod detection

xqrs:
See <https://docs.discord.food/reference#launch-signature>.

Co-Authored-By: ossenthusiast


internal/http/props.go
```diff
@@ -64,3 +64,20 @@ func getSuperProps() (string, error) {
 
 	return base64.StdEncoding.EncodeToString(raw), nil
 }
+
+func newLaunchSignature() string {
```
ossenthusiast: WDYT about renaming this to just generateLaunchSignature instead? Since {n,N}ew* is used for constructors in Go land.

xqrs: yeah

ayn2op (approved):
LGTM! Thank you <3

PR #692, opened 2026-01-09
feat(ui/chat): add inline image preview support

linuxmobile:
## Summary
Adds inline image previews for attachments directly in the terminal, with automatic protocol detection and comprehensive fallback support.

| Zellij (ANSI) | Foot (sixel) | Ghostty (kitty) |
|--------|------|---------|
| image | image | image |

## Features
- Support for Kitty, iTerm2, and Sixel graphics protocols
- ANSI half-block (▀) fallback for unsupported terminals
- Automatic multiplexer detection (tmux/zellij/screen) falls back to ANSI
- Configurable via `[image_previews]` in config.toml (enabled, type, max_height, max_width)
- WebP/GIF/JPEG/PNG format support

## Architecture
- New `internal/media` package with protocol detection and image rendering
- Marker-based rendering using reserved color ID range (0x100000+)
- Platform-specific terminal cell size detection (Unix/Windows)
- LRU cache with configurable eviction (100 images)

## Stability
- Context cancellation for downloads on channel switch
- Scroll debouncing (150ms) to prevent render thrashing
- Download size limit (25MB) and pixel limit (100MP) to prevent OOM
- Race-safe timer callbacks using generation counter

## Configuration
```toml
[image_previews]
enabled = false
type = "auto"  # auto, kitty, sixel, iterm, ansi
max_height = 10
max_width = 40

### CODE REVIEW

**Comment by siphc** [comment]
ANSI works on st + tmux-256color, lgtm

**Comment by ossenthusiast** [comment]
If the image is near the top instead of bottom, the image is not drawn at all and it displays a blank spot instead of the image. This is not the case if the image is near the bottom, however.

Actual image (works fine):
image

Near top (blank spot):
image

Near bottom (partially drawn, correct behavior):
image

File: internal/ui/chat/messages_list.go [hunk 1]
```diff
@@ -71,12 +116,97 @@ func newMessagesList(cfg *config.Config, chatView *View) *messagesList {
...
+
+	proto := media.DetectProtocol()
+	if proto == media.ProtoFallback {
+		return
+	}
+
+	x, y, w, h := ml.GetInnerRect()
+	if w <= 0 || h <= 0 {
+		return
+	}
```
ossenthusiast, line 154: Redundant check: *tview.Box already does this for you

ossenthusiast:
Use `(tcell.Screen).Tty` instead of writing to `os.Stdout`, you can also get rid of cellsize_*.go, since `tty.Tty` (`(tcell.Screen).Tty`) has a `(tty.Tty).WindowSize` method. You can use `(tty.WindowSize).CellDimensions` to get the cell size. It does the exact same thing:
```go
// CellDimensions returns the dimensions of a single cell, in pixels
func (ws WindowSize) CellDimensions() (int, int) {
	if ws.PixelWidth == 0 || ws.PixelHeight == 0 {
		return 0, 0
	}
	return (ws.PixelWidth / ws.Width), (ws.PixelHeight / ws.Height)
}
```

internal/media/cellsize_unix.go
```diff
@@ -0,0 +1,20 @@
...
+func getCellSizeFromTerminal() (int, int, bool) {
+	ws, err := unix.IoctlGetWinsize(int(os.Stdout.Fd()), unix.TIOCGWINSZ)
+	if err != nil {
+		return 0, 0, false
+	}
+	if ws.Col == 0 || ws.Row == 0 || ws.Xpixel == 0 || ws.Ypixel == 0 {
+		return 0, 0, false
+	}
+	return int(ws.Xpixel) / int(ws.Col), int(ws.Ypixel) / int(ws.Row), true
+}
```
ossenthusiast, line 20: Can be removed in favor of `(tty.WindowSize).CellDimensions`

internal/config/config.toml
```diff
@@ -35,6 +35,18 @@ duration = 0
 enabled = true
 only_on_ping = true
 
+[image_previews]
+# Enable inline image previews in messages
+enabled = false
```
ossenthusiast, line 40: Enable image previews by default. WDYT?

ossenthusiast, line 40:
```suggestion
enabled = true
```

internal/ui/chat/messages_list.go
ossenthusiast, line 154:
```suggestion
```

TechyTrickster:
hello!  poking my head in here.  What was the reason for abandoning this PR?  (or closing with unmerged changes)  It looks like this was pretty promising.

PR #711, opened 2026-01-16
fixed input not focusing on ctrl+i or ctrl+m

Wulgaren:


ayn2op:
If a user sets `keys.focus_message_input` as `"Ctrl+I"` (which is the default keymap for `focus_message_input`), your "fix" steals all `Tab` key events and turns them into `Ctrl+I` even when the user actually wants `Tab` to behave as `Tab`. Same for `Enter` when bound to `Ctrl+M`.

Wulgaren:
@ayn2op yeah im aware but i dont see any other solution to this. why is focusing input by default set to a keyboard shortcut that is interpreted as something else (tab)? or is this just broken on macos?

ayn2op:
> @ayn2op yeah im aware but i dont see any other solution to this. why is focusing input by default set to a keyboard shortcut that is interpreted as something else (tab)? or is this just broken on macos?

That’s not macOS-specific or broken, it’s how old terminals work at the ASCII level. In a terminal, Ctrl+key doesn’t mean “modifier + key”, it means “send a control character”. Historically:
- Ctrl+H = ASCII 8 = Backspace
- Ctrl+I = ASCII 9 = Tab

This dates back to teletype terminals (pre–Backspace key), and terminals still honors it for compatibility. The physical Backspace and Tab keys just emit those same control codes.

The real solution is to use a modern terminal that supports CSI-u so keys are reported as keys, not characters. With CSI-u enabled, Ctrl+I is not reported as Tab and same goes for Ctrl+H.

PR #712, opened 2026-01-16
added quick switcher

Wulgaren:
added quick switcher accessible via Ctrl + K

Screen Recording 2026-01-17 at 10 09 44 - 01


ayn2op:
Can you add a preview to the PR description? Also, instead of using an `InputField` for the quickswitcher, you could use a floating modal instead, check the `(*chat.View).showConfirmModal` for reference.

Something like this would be nice (neovim):
image
image

Wulgaren:
added the preview. i was inspired by the command window from vim for the design but the modal is also a really nice choice

Wulgaren:
added the modal

internal/ui/chat/picker.go
```diff
@@ -0,0 +1,241 @@
...
+
+	candidates []channelCandidate
+}
+
+func newQuickSwitcher(view *View, cfg *config.Config) *quickSwitcher {
+	qs := &quickSwitcher{
+		Flex: tview.NewFlex(),
+		view: view,
+		cfg:  cfg,
+	}
```
ossenthusiast, line 31: Instead of calling `ui.ConfigureBox` on the list, call it on the flex. Example: https://github.com/ayn2op/discordo/blob/main/internal/ui/chat/guilds_tree.go#L32

Make sure to pass a tview.NewBox() to `ConfigureBox`, the reason being that `tview.Flex` sets `(*tview.Box).dontClear` to `true`, which causes the content of the `tview.Primitive`s underneath to show up in the *floating modal* (tview.Flex).

internal/ui/chat/quick_switcher.go
```diff
@@ -0,0 +1,241 @@
...
+	qs.inputField.SetFieldBackgroundColor(tview.Styles.PrimitiveBackgroundColor)
+	qs.inputField.SetFieldTextColor(tview.Styles.PrimaryTextColor)
+	qs.inputField.SetLabel("Jump to: ")
+	qs.inputField.SetChangedFunc(qs.onInputChanged)
+	qs.inputField.SetDoneFunc(qs.onDone)
+	qs.inputField.SetInputCapture(qs.onInputFieldCapture)
+
+	// Create list for autocomplete suggestions with border
+	qs.list = tview.NewList()
+	qs.list.Box = ui.ConfigureBox(qs.list.Box, &cfg.Theme)
```
ossenthusiast: This can be removed: https://github.com/ayn2op/discordo/pull/712/changes#r2705996085

internal/ui/chat/picker.go
Wulgaren: done

internal/ui/chat/quick_switcher.go
```diff
@@ -30,6 +30,8 @@ func newQuickSwitcher(view *View, cfg *config.Config) *quickSwitcher {
 		cfg:  cfg,
 	}
 
+	qs.Flex.Box = ui.ConfigureBox(tview.NewBox(), &cfg.Theme)
```
ossenthusiast:
```diff
diff --git a/internal/ui/chat/quick_switcher.go b/internal/ui/chat/quick_switcher.go
index b10df94..c831dc4 100644
--- a/internal/ui/chat/quick_switcher.go
+++ b/internal/ui/chat/quick_switcher.go
@@ -30,7 +30,7 @@ func newQuickSwitcher(view *View, cfg *config.Config) *quickSwitcher {
 		cfg:  cfg,
 	}
 
-	qs.Flex.Box = ui.ConfigureBox(tview.NewBox(), &cfg.Theme)
+	qs.Box = ui.ConfigureBox(tview.NewBox(), &cfg.Theme)
 
 	// Create input field
 	qs.inputField = tview.NewInputField()
```

internal/ui/chat/quick_switcher.go
```diff
@@ -0,0 +1,241 @@
...
+	if index >= 0 && index < len(qs.candidates) {
+		candidate := qs.candidates[index]
+		qs.view.guildsTree.SelectChannelID(candidate.id)
+		qs.view.toggleQuickSwitcher()
+	}
+}
+
+func (qs *quickSwitcher) onListInputCapture(event *tcell.EventKey) *tcell.EventKey {
+	switch event.Key() {
+	case tcell.KeyEnter:
```
ossenthusiast: Instead of hardcoding keys here, you can have a `picker` table in the config.
```go
type PickerConfig struct {
    Cancel string `toml:"cancel"`
    Previous string `toml:"previous"`
    Next string `toml:"next"`
    Confirm string `toml:"confirm"`
}

Comment (ossenthusiast) [reply internal/ui/chat/quick_switcher.go]: These will be used here instead:
```go
switch event.Key() {
case qs.cfg.Keys.Picker.Cancel:
case qs.cfg.Keys.Picker.Confirm:
}
```

internal/ui/chat/quick_switcher.go
ossenthusiast: `OpenQuickSwitcher` can also be added here instead of `open_quick_switcher`:
```go
type PickerConfig struct {
Open string `toml:"open"`
// ...
}
```

internal/ui/chat/quick_switcher.go
```diff
@@ -0,0 +1,230 @@
...
+			id:        ch.ID,
+		})
+	}
+
+	var candidateStrings []string
+	for _, c := range candidates {
+		candidateStrings = append(candidateStrings, c.String())
+	}
+
+	matches := fuzzy.Find(currentText, candidateStrings)
```
ossenthusiast: Use `fuzzy.FindFrom` instead of collecting into `candidateStrings`: https://github.com/ayn2op/discordo/blob/main/internal/ui/chat/message_input.go#L318

internal/ui/chat/quick_switcher.go
```diff
@@ -0,0 +1,230 @@
...
+		cfg:  cfg,
+	}
+
+	qs.Flex.Box = ui.ConfigureBox(tview.NewBox(), &cfg.Theme)
+
+	// Create input field
+	qs.inputField = tview.NewInputField()
+	qs.inputField.SetFieldBackgroundColor(tview.Styles.PrimitiveBackgroundColor)
+	qs.inputField.SetFieldTextColor(tview.Styles.PrimaryTextColor)
+	qs.inputField.SetLabel("Jump to: ")
```
ossenthusiast: I think "Jump to: " should be changed to just "> "

internal/ui/chat/quick_switcher.go
```diff
@@ -0,0 +1,230 @@
...
+		view: view,
+		cfg:  cfg,
+	}
+
+	qs.Flex.Box = ui.ConfigureBox(tview.NewBox(), &cfg.Theme)
+
+	// Create input field
+	qs.inputField = tview.NewInputField()
+	qs.inputField.SetFieldBackgroundColor(tview.Styles.PrimitiveBackgroundColor)
+	qs.inputField.SetFieldTextColor(tview.Styles.PrimaryTextColor)
```
ossenthusiast: These are the defaults and can be removed.

internal/ui/chat/quick_switcher.go
```diff
@@ -0,0 +1,230 @@
...
+					name:      "#" + ch.Name,
+					guildName: guild.Name,
+					id:        ch.ID,
+				})
+			}
+		}
+	}
+
+	// DM Channels
+	privateChannels, _ := qs.view.state.PrivateChannels()
```
ossenthusiast: Handle err: https://github.com/ayn2op/discordo/blob/2e7cd26517593e220b34436c167cd85d55f4102f/internal/ui/chat/guilds_tree.go#L237

internal/ui/chat/quick_switcher.go
```diff
@@ -0,0 +1,230 @@
...
+		qs.list.Clear()
+		return
+	}
+
+	var candidates []channelCandidate
+
+	// Guild Channels
+	guilds, _ := qs.view.state.Cabinet.Guilds()
+	for _, guild := range guilds {
+		channels, _ := qs.view.state.Cabinet.Channels(guild.ID)
```
ossenthusiast: Handle err: https://github.com/ayn2op/discordo/blob/2e7cd26517593e220b34436c167cd85d55f4102f/internal/ui/chat/guilds_tree.go#L156

internal/ui/chat/quick_switcher.go
ossenthusiast: https://github.com/ayn2op/discordo/blob/2e7cd26517593e220b34436c167cd85d55f4102f/internal/ui/chat/guilds_tree.go#L157C2-L160C1

internal/ui/chat/guilds_tree.go
```diff
@@ -313,3 +312,99 @@ func (gt *guildsTree) yankID() {
 		go clipboard.Write(clipboard.FmtText, []byte(id.String()))
 	}
 }
+
+func (gt *guildsTree) SelectChannelID(channelID discord.ChannelID) {
+	if gt.chatView.state == nil || gt.chatView.state.Cabinet == nil {
+		return
+	}
```
ossenthusiast: Redundant check

ossenthusiast: state cannot be nil at this point

internal/ui/chat/quick_switcher.go
ossenthusiast: PickerKeys* or QuickSwitcherKeys* in keys.go

internal/ui/chat/quick_switcher.go
ossenthusiast: Clarification: NewInputField already sets both of these to `Styles.PrimitiveBackgroundColor` and `Styles.PrimaryTextColor`, respectively.

```go
func NewInputField() *InputField {
// ...
	i.textArea.textStyle = tcell.StyleDefault.Background(Styles.ContrastBackgroundColor).Foreground(Styles.PrimaryTextColor)
	i.textArea.placeholderStyle = tcell.StyleDefault.Background(Styles.ContrastBackgroundColor).Foreground(Styles.ContrastSecondaryTextColor)
// ...
}
```

internal/ui/chat/quick_switcher.go
```diff
@@ -0,0 +1,230 @@
...
+	qs.inputField.SetLabel("Jump to: ")
+	qs.inputField.SetChangedFunc(qs.onInputChanged)
+	qs.inputField.SetDoneFunc(qs.onDone)
+	qs.inputField.SetInputCapture(qs.onInputFieldCapture)
+
+	// Create list for autocomplete suggestions
+	qs.list = tview.NewList()
+	qs.list.ShowSecondaryText(false)
+	qs.list.SetSelectedFunc(qs.onListSelected)
+	qs.list.SetInputCapture(qs.onListInputCapture)
```
ossenthusiast: Instead of adding an input capture to the list as well and switching focus between them to navigate and select items, you could forward key events from `*tview.InputField`'s input capture to the `*tview.List`'s input handler, so the end user does not have to switch focus for navigation. For instance, `Ctrl+P` and `Ctrl+N` can be used to select between items in the list instead.

Example (in input capture of InputField):
```go
qs.list.InputHandler()(tcell.NewEventKey(KeyUp, "", tcell.ModNone), nil)
return nil
```

This will forward the `KeyUp` key event to the internal list, `*tview.List` handles `KeyUp` events internally by select the previous item in the list.

ossenthusiast: This is also done for mentionsList, check `message_input.go`:
https://github.com/ayn2op/discordo/blob/main/internal/ui/chat/message_input.go#L154

Wulgaren:
done!

internal/config/keys.go
```diff
@@ -53,6 +53,14 @@ type (
 		SelectionKeys
 	}
 
+	PickerKeys struct {
+		Open    string `toml:"open"`
+		Cancel  string `toml:"cancel"`
+		Up      string `toml:"up"`
+		Down    string `toml:"down"`
```
ossenthusiast, line 60: Just embed `SelectionKeys` in this struct and forward key events for all of the keys in `SelectionKeys` to the internal list.

internal/config/keys.go
```diff
@@ -65,6 +73,7 @@ type (
 		MessagesList MessagesListKeys `toml:"messages_list"`
 		MessageInput MessageInputKeys `toml:"message_input"`
 		MentionsList MentionsListKeys `toml:"mentions_list"`
+		Picker       PickerKeys        `toml:"picker"`
```
ossenthusiast, line 76: QuickSwitcher or Picker? I prefer picker, since we might decide to have multiple pickers in future. WDYT?

internal/ui/chat/guilds_tree.go
```diff
@@ -313,3 +313,99 @@ func (gt *guildsTree) yankID() {
 		go clipboard.Write(clipboard.FmtText, []byte(id.String()))
 	}
 }
+
+func (gt *guildsTree) SelectChannelID(channelID discord.ChannelID) {
+	if gt.chatView.state.Cabinet == nil {
+		return
+	}
```
ossenthusiast, line 320: Redundant

internal/ui/chat/picker.go
```diff
@@ -0,0 +1,207 @@
...
+			}
+			return nil
+		}
+	}
+
+	switch event.Name() {
+	case qs.cfg.Keys.Picker.Cancel:
+		qs.view.toggleQuickSwitcher()
+		return nil
+	}
```
ossenthusiast, line 193: Why separate statement? This can be merged with the previous switch.

internal/ui/chat/quick_switcher.go
```diff
@@ -0,0 +1,207 @@
...
+	switch key {
+	case tcell.KeyEnter:
+		// If there are candidates, select the first one
+		if len(qs.candidates) > 0 {
+			candidate := qs.candidates[0]
+			qs.view.guildsTree.SelectChannelID(candidate.id)
+			qs.view.toggleQuickSwitcher()
+		}
+	}
+}
```
ossenthusiast: This can be merged with the `Confirm` keymap handler in the InputField input capture.

internal/ui/chat/quick_switcher.go
```diff
@@ -0,0 +1,207 @@
...
+	qs.inputField = tview.NewInputField()
+	qs.inputField.SetLabel("> ")
+	qs.inputField.SetChangedFunc(qs.onInputChanged)
+	qs.inputField.SetDoneFunc(qs.onDone)
+	qs.inputField.SetInputCapture(qs.onInputFieldCapture)
+
+	// Create list for autocomplete suggestions
+	qs.list = tview.NewList()
+	qs.list.ShowSecondaryText(false)
+	qs.list.SetSelectedFunc(qs.onListSelected)
```
ossenthusiast: Redundant since we are already handling `Confirm` key in the input capture.

internal/config/keys.go
Wulgaren: i was thinking about quick switcher bcuz iirc its the name used by discord, but yeah i think picker is good if we want a reusable component!

ayn2op:
#723

PR #713, opened 2026-01-16
added infinite scrolling for old messages

Wulgaren:


Z1xus:
this is very broken for me. probably because of a race condition of some sorts it loads the messagfes in even after switching the channel which makes it display messages from the last chjannel in the current one.

Wulgaren:
i fixed the race condition but theres a problem while scrolling with mouse, im not sure how to resolve

ayn2op:
https://github.com/ayn2op/discordo/pull/716

PR #716, opened 2026-01-16
feat(ui/chat): fetch older messages on select-scrolling

### LINKED ISSUE #308: Fetch more messages by scrolling
Currently when selecting a channel and scrolling through its history, there is no way of fetching messages beyond the `messages_limit` messages loaded in at the time of selection. A way of querying for older messages, either implicitly by trying to scroll or an explicit keypress to trigger it would be nice. The discord API *does* allow [message fetching with an offset](https://pkg.go.dev/github.com/diamondburned/arikawa/v3@v3.3.0/api#Client.MessagesBefore) so those are some good news.

A temporary fix for now is obviously to just increase the messages limit, but this of course is suboptimal when you really only *want* to preview few messages as you're now stuck with slower channel loading times with no real benefit.

ayn2op:
Closes #308

### DIFF
--- a/internal/config/config.toml
+++ b/internal/config/config.toml
@@ -19,7 +19,7 @@ show_attachment_links = true
 # Note: Use autocomplete_limit = 0 to disable.
 autocomplete_limit = 20
 
-# The number of messages to fetch when a text-based channel is selected from guilds tree. The minimum and maximum value is 0 and 100, respectively.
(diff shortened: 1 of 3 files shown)


ayn2op:
Preview:
https://github.com/user-attachments/assets/e3026f1c-b8b7-4459-9117-050e719df244

Wulgaren:
there’s an error that duplicates some of the messages upon scrolling, mouse scrolling doesn’t work too

PR #727, opened 2026-01-27
feat(messages_list): Add spoiler support

### LINKED ISSUE #138: Feature request: Spoiler tags
I think it would just be a pretty useful feature, by default, spoilered (surrounded by unescaped `||`) contents of a message could be written as `█` (U+2588) and pressing enter while hovering over a message would reveal its contents. Additionally, an option to automatically display spoilered contents like in the official discord client would be neat.

xqrs:
ningen/discordmd bug: the parsing of spoilers is inline, but it's actually a cross-block flag. Most use cases will work without any issues.

Fixes #138


xqrs:
out
(the extra newline is not a bug, it's a heading, it has a hard newline after it)

internal/markdown/renderer.go
```diff
@@ -73,22 +77,26 @@ func (r *Renderer) renderFencedCodeBlock(w io.Writer, node *ast.FencedCodeBlock,
...
+			io.WriteString(w, r.checkAndSpoil("| "))
+			w.Write(r.checkAndSpoilSource(line.Value(source)))
 		}
 	}
 }
 
 func (r *Renderer) renderAutoLink(w io.Writer, node *ast.AutoLink, entering bool, source []byte) {
+	if r.spoil {
+		return
+	}
```
ossenthusiast: Is this consistent with official client? Does it not render links if the content is being is spoiled? If so, I would recommend adding a comment on top.

xqrs: ningen/discordmd won't even parse it correctly, that's a skill issue from their side. (and if it's parsed correctly, it will be consistent with the official client)

juice928:
👋 Hi, I'm an automated AI code review bot. I ran some checks on this PR and found **3 points that might be worth attention** (could be false positives, please use your judgment):

1. <strong>Logic adjustment for nested spoilers</strong>
   - [Location](https://github.com/ayn2op/discordo/blob/738d103f89874b84e9df4b4f8dc40fe3352d79da/internal/markdown/renderer.go#L170-L174): `internal/markdown/renderer.go:L170-L174`
   - Impact: When spoilers are nested, the outer content might be revealed prematurely after exiting an inner block.
   - Suggestion: Consider using a numeric counter to track nesting levels instead of a simple boolean flag.

2. <strong>Spoiler length calculation for multi-byte characters</strong>
   - [Location](https://github.com/ayn2op/discordo/blob/738d103f89874b84e9df4b4f8dc40fe3352d79da/internal/markdown/renderer.go#L239-L251): `internal/markdown/renderer.go:L239-L251`
   - Impact: The masking length might appear incorrect for multi-byte characters like emojis or CJK text.
   - Suggestion: Using rune counting would ensure the number of spoiler blocks matches the visual character count.

3. <strong>Refining the default spoiler configuration behavior</strong>
   - [Location](https://github.com/ayn2op/discordo/blob/738d103f89874b84e9df4b4f8dc40fe3352d79da/internal/ui/chat/messages_list.go#L136-L138): `internal/ui/chat/messages_list.go:L136-L138`
   - Impact: Defaulting to "show" for unknown configuration values might unintentionally expose hidden content.
   - Suggestion: Consider defaulting to "hide" to ensure a safer, privacy-first experience for users.

If you find these suggestions disruptive, you can reply "stop" , and I'll automatically skip this repository in the future.

PR #730, opened 2026-02-01
cleanup(ui/chat): Cache @me in chatView

xqrs:


ayn2op:
regarding #730, i dont think caching the client user (me) on our side is the way to go, the reason being that the cabinets are updated in various state events. currently for `store.MeStore`, it is updated in only two events—`*gateway.ReadyEvent` and `*gateway.UserUpdateEvent`, but this may change in future. in arikawa, the state package is effectively a materialized view of gateway + http api—the state updates its internal caches (`(state.State).Cabinet`/store) based on those events, and if something is missing it can fall back to the api and then store the result (`(state.State).Me` vs `(state.State).Cabinet.Me`). furthermore, the library even calls out the idea of a pluggable cache (`store.MeStore`, ...) and api fallback as a core design point. besides, `(defaultstore.MeStore).Me` already pulls it from cache, it is not making any http requests:

```go
type Me struct {
    mut  sync.RWMutex
    self discord.User
}
```

the correct approach would be to just ignore the error. a helper might be okay too:

```go
func (v *View) getMe() *discord.User { /* ignore err and return me */ }
```
examples: 
- <https://github.com/search?q=repo%3Adiamondburned%2Fdissent%20Me()&type=code>
- <https://github.com/diamondburned/dissent/blob/200243461ab946e744596818f902d78ee379485f/internal/messages/message.go#L185>

on an unrelated note, moving processText and expandMentions to the messageInput struct makes sense

PR #743, opened 2026-02-11
feat(ui): Allow arrow keys and Home/End to be always active

xqrs:
Non-QWERTY users will get a headache by our default configuration. We should always allow the cursor+Home/PgUp/PgDn/End keys to function.


ayn2op:
This hard-codes those keys in the input handler, which can cause a lot of confusion, since the hard-coded keybinds can no longer be bound to something else.

PR #744, opened 2026-02-12
fix(config): handle EDITOR that may pass flags on *nix

minneelyyyy:
This fixes CTRL+E not working if your EDITOR has spaces in it, such as for flags in `emacs -nw`


xqrs:
Technically speaking, the `$EDITOR` environment variable should be treated as a raw string for the shell.

`sh -c "$EDITOR myfile"` is the correct thing todo.

internal/ui/chat/message_input.go
```diff
@@ -614,7 +614,16 @@ func (mi *messageInput) editor() {
...
+	editorArgs := strings.Split(mi.cfg.Editor, " ")
+	if len(editorArgs) == 0 {
+		return
+	}
+
+	editor := editorArgs[0]
+	args := append(editorArgs[1:], file.Name())
+
+	cmd := exec.Command(editor, args...)
+
```
xqrs: this should be:
```go
cmd := exec.Command("sh", "-c", ml.cfg.Editor + " " + file.Name())
```

internal/ui/chat/message_input.go
```diff
@@ -614,7 +614,16 @@ func (mi *messageInput) editor() {
 
 	file.WriteString(mi.GetText())
 
-	cmd := exec.Command(mi.cfg.Editor, file.Name())
+	editorArgs := strings.Split(mi.cfg.Editor, " ")
```
ayn2op: use `strings.Fields` here instead: https://pkg.go.dev/strings#Fields

internal/ui/chat/message_input.go
minneelyyyy: fixed

internal/ui/chat/message_input.go
```diff
@@ -614,7 +614,8 @@ func (mi *messageInput) editor() {
 
 	file.WriteString(mi.GetText())
 
-	cmd := exec.Command(mi.cfg.Editor, file.Name())
+	cmd := exec.Command("sh", "-c", mi.cfg.Editor+" "+file.Name())
```
ayn2op: `sh` is not cross-platform.

xqrs: we do a simple platform check.
```go
if runtime.GOOS == "windows" {
  // windows stuff
} else {
  // default
}
```

xqrs:
```go
var cmd *exec.Cmd
if runtime.GOOS == "windows" {
    cmd = exec.Command(mi.cfg.Editor, file.Name()) // Windows, do things normally
} else {
    // provide the filename as a proper argument so it won't break if it has spaces.
    // e.g., {"sh", "-c", "myeditor \"$1\"", "--", "My FileName"} runs "myeditor" "My FileName"
    cmd = exec.Command("sh", "-c", mi.cfg.Editor + " \"$1\"", "--", file.Name())
}
```

minneelyyyy: done

internal/ui/chat/message_input.go
```diff
@@ -604,6 +605,11 @@ func (mi *messageInput) stopTabCompletion() {
 }
 
 func (mi *messageInput) editor() {
+	if mi.cfg.Editor == "" {
+		slog.Warn("Attempt to open editor, but no editor is configured. Configure one in your config.toml or ensure that EDITOR is set in your environment.");
```
minneelyyyy: could go in config.go instead and warn when attempting to initialize `mi.cfg.Editor`.

minneelyyyy:
ready for reviews :3

xqrs:
looks good, but I think we overcomplicated it a little bit

minneelyyyy:
> looks good, but I think we overcomplicated it a little bit

Probably, but it’s likely the best way to check for a Unix-like. I could revert to the match if necessary.

xqrs (dismissed):
LGTM

internal/config/config.go
```diff
@@ -139,3 +140,13 @@ func applyDefaults(cfg *Config) {
...
+
+// Open file `path` in the user's configured editor
+func (cfg *Config) OpenEditor(path string) *exec.Cmd {
+	if cfg.Editor == "" {
+		slog.Warn("Attempt to open file with editor, but no editor is set")
+		return nil
+	}
+
+	return cfg.createEditorCommand(path)
+}
```
ayn2op: I would prefer inlining this function instead.

minneelyyyy: How so? As in don't have this function at all?

ayn2op:
This also breaks when there are spaces in the editor path.

minneelyyyy:
> This also breaks when there are spaces in the editor path.

that's not possible to support reasonably using `EDITOR`, that is just user error. The point of this fix is to support flags passed to your editor, which requires interpreting `EDITOR` as a shell command. There is precedent for this solution in the Git source code:
```c
static const char **prepare_shell_cmd(struct strvec *out, const char **argv)
{
	if (!argv[0])
		BUG("shell command is empty");

	if (strcspn(argv[0], "|&;<>()$`\\\"' \t\n*?[#~=%") != strlen(argv[0])) {
		strvec_push_nodup(out, git_shell_path());
		strvec_push(out, "-c");

		/*
		 * If we have no extra arguments, we do not even need to
		 * bother with the "$@" magic.
		 */
		if (!argv[1])
			strvec_push(out, argv[0]);
		else
			strvec_pushf(out, "%s \"$@\"", argv[0]);
	}

	strvec_pushv(out, argv);
	return out->v;
}
```
```
$ export GIT_EDITOR='/home/mins/editor test' 
$ git rebase -i HEAD~2
hint: Waiting for your editor to close the file... /home/mins/editor test: line 1: /home/mins/editor: No such file or directory
error: there was a problem with the editor '/home/mins/editor test'
```

internal/config/config.go
ayn2op: > How so? As in don't have this function at all?

Yes

ayn2op (approved):
Thank you! ❤️

PR #745, opened 2026-02-12
feat(ui/chat): add syntax highlighting in codeblocks

ayn2op:
Preview:
Screenshot 2026-02-12 at 6 33 08 PM


xqrs:
What if the user disabled syntax highlighting?
How about a `code_block_style` config option with different background from the default message background, just like discord's way of showing unhighlighted codeblocks

ayn2op:
I have already addressed this with explicit fenced code-block headers, so unhighlighted blocks are still distinguishable from normal text:

- code for fences without a language.
- code: <language> for unknown language tags.
- code: analyzed when language is analyzed from content.

PR #746, opened 2026-02-12
feat(messages_list): wrap code blocks correctly

xqrs:
depends on [tview#8](https://github.com/ayn2op/tview/pull/8)


xqrs:
updated

PR #748, opened 2026-02-13
feat(workflows): add goreleaser

stauersbol:
Simply adds goreleaser, there is a bunch of defaults and some things are locked behind their pro plan, but might not be necessarily useful things.

There could be some debate on the `.goreleaser.yaml` config file, but I'm all ears on that part.
It defaults to prerelease as this was a request from ayyan.

Feel free to go read through the docs at: https://goreleaser.com/customization/


.goreleaser.yaml
```diff
@@ -0,0 +1,12 @@
+project_name: discordo
```
ayn2op:
```suggestion
# yaml-language-server: $schema=https://goreleaser.com/static/schema.json
project_name: discordo
```
https://goreleaser.com/customization/#json-schema

.goreleaser.yaml
```diff
@@ -0,0 +1,12 @@
...
+  - env: [CGO_ENABLED=1]
+    goos:
+      - linux
+      - windows
+      - darwin
+    goarch:
+      - amd64
+      - arm64
+release:
+  prerelease: "true"
```
ayn2op: I think it might be better to do a proper release instead of prerelease.

.goreleaser.yaml
```diff
@@ -0,0 +1,12 @@
+project_name: discordo
+builds:
+  - env: [CGO_ENABLED=1]
+    goos:
+      - linux
+      - windows
+      - darwin
+    goarch:
+      - amd64
+      - arm64
```
ayn2op, line 20: Use the default goarch (amd64, arm64, 386) instead of manually specifying.

.goreleaser.yaml
ayn2op: Add a `go mod tidy` `before` hook: https://goreleaser.com/customization/builds/go/?h=goarch#go-modules

```yaml
before:
  hooks:
    - go mod tidy
```

.goreleaser.yaml
stauersbol: alright

.goreleaser.yaml
stauersbol: corrected

.goreleaser.yaml
stauersbol: added

stauersbol:
looks like i forgot to setup signed commits, will have to do that

stauersbol:
should all be signed now after amending them

ayn2op:
https://github.com/goreleaser/goreleaser-cross

stauersbol:
> [goreleaser/goreleaser-cross](https://github.com/goreleaser/goreleaser-cross?rgh-link-date=2026-02-16T23%3A54%3A09Z)

yep that was the plan to use, didnt have time to look into it yesterday, plan is doing so day.

stauersbol:
@ayn2op need you opinion on some bits, ive added comments to what im unsure about. for the most part dry-run works.

I believe there is still some configuration left and I might just nuke the custom name templating I did, as it makes the life of packagers easier on AUR and such.

stauersbol:
I might also consider adding linux arm64 and arm64hf (in case pi users would like to use it)

.goreleaser.yaml
```diff
@@ -0,0 +1,59 @@
+# yaml-language-server: $schema=https://goreleaser.com/static/schema.json
+project_name: discordo
```
ayn2op: `project_name` is auto-inferred (can be removed): https://goreleaser.com/customization/project/?h=project_name

.goreleaser.yaml
```diff
@@ -0,0 +1,59 @@
...
+# NOTE: if this is not here LSPs will complain about missing version
+# (only relevant for goreleaser, does not impact actual version tags)
+version: 2
+env:
+  - CGO_ENABLED=1
+
```
ayn2op, line 10: Add `go mod tidy` before hook.

```yaml
before:
  hooks:
    - go mod tidy
```

.goreleaser.yaml
```diff
@@ -0,0 +1,59 @@
...
+    goos: [darwin]
+    goarch: [arm64]
+    env:
+      - CGO_ENABLED=1
+      - CC=oa64-clang
+      - CXX=oa64-clang++
+
+archives:
+  # NOTE: assuming binary and tar.gz is good enough for the time being.
+  - formats: ["binary", "tar.gz"]
```
ayn2op:
```suggestion
  - formats: ["zip", "tar.gz"]
```

.goreleaser.yaml
```diff
@@ -0,0 +1,59 @@
...
+    # NOTE: keep default naming or use custom? Default includes version, so it might help AUR packagers and others?
+    name_template: >-
+      discordo_
+      {{- if eq .Os "darwin" }}macos{{ else }}{{ .Os }}{{ end }}_
+      {{- if eq .Arch "amd64" }}x86_64
+      {{- else if eq .Arch "arm64" }}arm64
+      {{- else }}{{ .Arch }}{{ end }}
```
ayn2op: Default should be enough.


```suggestion
```

.goreleaser.yaml
```diff
@@ -0,0 +1,59 @@
...
+
+changelog:
+  # NOTE: do we disable changelog? It can be customized, so that it includes the author and such.
+  disable: true
+
+before:
+  hooks:
+    - go mod tidy
```
ayn2op: This should be moved to the top, below version field. Check the default configuration using `goreleaser init`.

.github/workflows/go-releaser.yml
```diff
@@ -0,0 +1,32 @@
+# .github/workflows/go-releaser.yml
+name: goreleaser
+
+on:
+  pull_request:
```
ayn2op: I don't think a release should be made on every PR 😅

.goreleaser.yaml
```diff
@@ -0,0 +1,59 @@
...
+    name_template: >-
+      discordo_
+      {{- if eq .Os "darwin" }}macos{{ else }}{{ .Os }}{{ end }}_
+      {{- if eq .Arch "amd64" }}x86_64
+      {{- else if eq .Arch "arm64" }}arm64
+      {{- else }}{{ .Arch }}{{ end }}
+
+changelog:
+  # NOTE: do we disable changelog? It can be customized, so that it includes the author and such.
+  disable: true
```
ayn2op: Not required? https://goreleaser.com/customization/changelog/?h=changelog

.goreleaser.yaml
```diff
@@ -0,0 +1,59 @@
+# yaml-language-server: $schema=https://goreleaser.com/static/schema.json
+project_name: discordo
+# NOTE: if this is not here LSPs will complain about missing version
+# (only relevant for goreleaser, does not impact actual version tags)
+version: 2
+env:
+  - CGO_ENABLED=1
```
ayn2op: Since each build sets CGO_ENABLED=1, this can be removed?

.github/workflows/go-releaser.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/setup-go@v5
+        with:
+          go-version: stable
+
+      - name: setup release environment
+        run: |-
+          echo 'GITHUB_TOKEN=${{secrets.GITHUB_TOKEN}}' > .release-env
+
+      - name: Release publish
+        run: make release
```
ayn2op:
```yaml
on:
  push:
    tags:
      - "*"

jobs:
  release:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v6

      - uses: actions/setup-go@v6
        with:
          go-version: stable

      - name: setup release environment
        run: |-
          echo 'GITHUB_TOKEN=${{secrets.GITHUB_TOKEN}}' > .release-env
      - name: release publish
        run: make release
```

stauersbol:
alright touched the points you had @ayn2op, i also went ahead to add linux arm64 and armv7 build targets, since it could be relevant if some users of those 2 ISAs appear.

stauersbol:
Though there is something with how changelog works, unsure how it works under CI conditions, but it may overwrite any auto generated changelog notes on a release, but it might be worth the initial try and then adjust if it does indeed overwrite

stauersbol:
simplified the build targets in the config and also added some docs in the readme that describes goreleaser

stauersbol:
Alright finally got time to clean up a little. I removed a merge commit, as they tend to not be good for git history.

I have removed the `Dockerfile` as that is not needed as we are now purego(cgo free) with a recent update to a dependency.

@ayn2op ready for whenever you choose to look at this again, no rush on your end of cource :)

EDIT: note to a potential nit, I left a static variable in the Makefile as it might be relevant if we chose, before merging this, to just set a version of goreleaser-cross and then update with small PRs down the line, open for both, left it on `latest` as that is what I had it set to in the `Dockerfile`.

.goreleaser.yaml
```diff
@@ -0,0 +1,61 @@
+# yaml-language-server: $schema=https://goreleaser.com/static/schema.json
+# NOTE: if this is not here LSPs will complain about missing version
+# (only relevant for goreleaser, does not impact actual version tags)
+version: 2
+
+before:
+  hooks:
+    - go mod tidy
```
ayn2op, line 8: Can we include `go generate ./...` in the goreleaser `before` hooks?

```yaml
before:
  hooks:
    - go mod tidy
    - go generate ./...
```

This keeps release artifacts consistent with the generated sources that the normal build expects, instead of relying on generated files already being up to date in the checkout.

This was not the case when this PR was initially opened.

.goreleaser.yaml
```diff
@@ -0,0 +1,61 @@
...
+      - goos: darwin
+        goarch: amd64
+        env:
+          - CC=o64-clang
+          - CXX=o64-clang++
+      - goarch: arm64
+        goos: darwin
+        env:
+          - CC=oa64-clang
+          - CXX=oa64-clang++
```
ayn2op: CGO is no longer required. The entire builds section can be replaced with
```yaml
builds:
  - main: ./...
    env:
      - CGO_ENABLED=0
    goos:
      - linux
      - windows
      - darwin
    goarch:
      - arm64
      - amd64
      - "386"
```

ayn2op: Furthermore, we also do not need the Makefile anymore.

stauersbol:
@ayn2op im guessing we still need to use `goreleaser-cross`, but the makefile as you pointed out is not needed anymore. I have removed that and cleaned up all the cgo magic stuff as the dep that required it is now purego, but idk if there is a slight security issue with how I'm currently passing in the `GITHUB_TOKEN` to the `docker run` command? If there is then I'm more than happy for feedback on that part

stauersbol:
fixed a merge conflict

.github/workflows/go-releaser.yml
```diff
@@ -0,0 +1,25 @@
...
+      - name: release publish
+        run: |
+          docker run \
+          --rm \
+          -e 'GITHUB_TOKEN=${{secrets.GITHUB_TOKEN}}' \
+          -v /var/run/docker.sock:/var/run/docker.sock \
+          -v `pwd`:/go/src/github.com/ayn2op/discordo \
+          -w /go/src/github.com/ayn2op/discordo \
+          ghcr.io/goreleaser-cross:latest \
+          release --clean
```
ayn2op, line 25: Like previously stated, discordo no longer relies on CGO to compile, so goreleaser-cross setup can be completely removed.

stauersbol: ah woops my bad, will adjust later today

ayn2op: I'd recommend running `goreleaser init` again and going from there.

PR #753, opened 2026-02-25
fix(ui): recover gracefully from auth failure

### LINKED ISSUE #685: Login screen bugging out
Likely this happenes when I launch discordo, building from source or downloading prebuilt changes nothing. Tried this and this, and it likely happenes only to login screen, when I paste token like [this](https://github.com/ayn2op/discordo#token) part says, everything works right. Everything about OS is in the footage itself.
https://github.com/user-attachments/assets/5f67846c-c655-4820-9c58-524f37749425

stavxyz:
## Summary

Fixes #685.

- When a saved token fails authentication (e.g. wrong token entered on first run), the app now clears the bad token from the keyring and falls back to the login form with an error modal showing what went wrong, instead of leaving the terminal in a broken state (blank screen, unresponsive to keystrokes)
- Also handles the same failure path when entering a bad token directly from the login form — the error is shown to the user instead of silently logged

## Test plan

- [ ] Build from source and run with no saved token — login form should appear normally
- [ ] Enter an invalid token and click Login — error modal should appear with the auth failure reason, login form remains usable
- [ ] Kill the app and relaunch — login form should appear (bad token was cleared from keyring)

### DIFF
--- a/internal/ui/login/form.go
+++ b/internal/ui/login/form.go
@@ -51,7 +51,7 @@ func NewForm(app *tview.Application, cfg *config.Config, done DoneFn) *Form {
 func (f *Form) login() {
 	token := f.form.GetFormItem(0).(*tview.InputField).GetText()
 	if token == "" {
-		f.onError(errors.New("token required"))
+		f.ShowError(errors.New("token required"))
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


Z1xus:
hey, just a heads up i actually have an open pr for the same issue (#700) from january that covers the terminal cleanup and the keyring logic

PR #755, opened 2026-02-26
fix(mentions_list): Use usernames, not display names

xqrs:
If a user with display name: `Ken` and username `corren_`, typing `@Ke<TAB>` should expand to `@corren_`, not `@Ken`, similar to the official client's behavior.


ayn2op (approved):
Thank you! <3

PR #756, opened 2026-02-26
fix(ui/chat): mark channel as read after fetching messages

### LINKED ISSUE #666: When a chat is selected it doesn't seem to be marked as read
When I reopen discordo the chat is marked as unread and when I open the discord webapp after selecting a chat it isn't marked as read there either.

Edit: Okay I've found out that seems to be only the case if there are A LOT of unread messages in a channel.

kenperkins:
## Summary
- Fixes #666
- `MarkRead` was called before `Messages()` populated the cabinet, causing ningen's [`markRead`](https://github.com/diamondburned/ningen/blob/98fbd92e134d/states/read/read.go#L204-L208) to early-return when it couldn't find the message in the store
- No ack was sent to Discord and no `UpdateEvent` fired, so the channel's unread/mention styling never cleared
- Fix: move `MarkRead` to after `Messages()` so the message is in the cabinet

## Test plan
- [x] Select an unread channel (white bold) — styling changes to dim (read)
- [x] Verified via debug logging that `channel.LastMessageID` matches the newest fetched message
- [x] Verified via debug logging that the cabinet lookup fails before `Messages()` and succeeds after

### DIFF
--- a/internal/ui/chat/guilds_tree.go
+++ b/internal/ui/chat/guilds_tree.go
@@ -314,14 +314,14 @@ func (gt *guildsTree) onSelected(node *tview.TreeNode) {
 			return
 		}
 
-		go gt.chatView.state.ReadState.MarkRead(channel.ID, channel.LastMessageID)
-
 		limit := gt.cfg.MessagesLimit
 		messages, err := gt.chatView.state.Messages(channel.ID, uint(limit))
 		if err != nil {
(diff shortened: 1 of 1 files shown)


kenperkins:
I tested this locally and saw the behavior I expected when selecting a white or underlined channel.

ayn2op (approved):

PR #757, opened 2026-02-26
feat(ui): display friend nicknames in DM channel list

kenperkins:
## Summary
- Bump ningen to pick up the new `FullRelationship()` method ([diamondburned/ningen#10](https://github.com/diamondburned/ningen/pull/10))
- Use friend nicknames (when set) in place of display names for 1:1 DMs in the guilds tree, channel picker, and message list title

## Details
Discord's friend nickname feature allows users to set per-friend nicknames visible only to themselves. Previously, ningen's relationship state discarded this field. With the upstream fix merged, we can now display these nicknames in `ChannelToString`.

The `ChannelToString` signature gains a `*ningen.State` parameter to enable the nickname lookup. All three call sites are updated.

## Test plan
- [x] Build: `go build ./...`
- [x] Run discordo, expand Direct Messages — friends with nicknames set show their nickname instead of display name
- [x] Channel picker and message list title also reflect the nickname

CleanShot 2026-02-26 at 15 17 23

### DIFF
--- a/internal/ui/util.go
+++ b/internal/ui/util.go
@@ -8,6 +8,7 @@ import (
 	"github.com/ayn2op/discordo/internal/config"
 	"github.com/ayn2op/tview"
 	"github.com/diamondburned/arikawa/v3/discord"
+	"github.com/diamondburned/ningen/v3"
 	"github.com/gdamore/tcell/v3"
 )
 
@@ -61,7 +62,7 @@ func Centered(p tview.Primitive, width, height int) tview.Primitive {
 		AddItem(p, 1, 1, 1, 1, 0, 0, true)
 }
 
-func ChannelToString(channel discord.Channel, icons config.Icons) string {
+func ChannelToString(channel discord.Channel, icons config.Icons, state *ningen.State) string {
 	var icon string
 	switch channel.Type {
 	case discord.DirectMessage, discord.GroupDM:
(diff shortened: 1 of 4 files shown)


internal/ui/util.go
```diff
@@ -71,6 +72,14 @@ func ChannelToString(channel discord.Channel, icons config.Icons) string {
 
 		recipients := make([]string, len(channel.DMRecipients))
 		for i, r := range channel.DMRecipients {
+			if state != nil && channel.Type == discord.DirectMessage {
+				if rel, ok := state.RelationshipState.FullRelationship(r.ID); ok {
```
ossenthusiast: What about retrieving the relationship struct only for RelationshipType == Friend?

kenperkins: Not unreasonable, to save on memory or just keep it only when needed?

kenperkins: Looking at this a bit more, the problem is this is all handled in `ningen`. I'm not against the idea, but we'd need to go back and change it there (again), unless I'm missing something.

kenperkins: @diamondburned this is a bit more in your side of the stack, do you have an opinion here?

diamondburned:
```go
if state != nil {
    for _, r := range channel.DMRecipients {
        rel, ok := state.RelationshipState.FullRelationship(r.ID)
        if !ok || rel.Type != discord.Friend {
            continue
        }
    }
}
```

probably

ossenthusiast: Yeah that should work

ayn2op (approved):

PR #759, opened 2026-02-27
fix(config): reset style instead of merging for StyleWrapper

xqrs:
- **fix(markdown/renderer): Don't force bold on mentions**

### DIFF
--- a/internal/config/config.toml
+++ b/internal/config/config.toml
@@ -169,7 +169,7 @@ down = "ctrl+n"
 top = "home"
 bottom = "end"
 
-# style = { foreground = "", background = "", attributes = "" or [""] }
+# style = { foreground = "", background = "", attributes = "" or ["",""],  underline = "", underline_color = "" }
 [theme.title]
 alignment = "left"                                           # `"left"`, `"center"`, or `"right"`.
 normal_style = { attributes = "dim" }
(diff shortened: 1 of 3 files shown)


internal/config/theme.go
```diff
@@ -41,48 +41,71 @@ func (sw *StyleWrapper) UnmarshalTOML(v any) error {
 		return errInvalidType
 	}
 
+	// Reset on new styles
+	sw.Style = tcell.StyleDefault
```
ayn2op, line 45: Could you please explain this change and what it does?

internal/config/config.toml
```diff
@@ -169,7 +169,7 @@ down = "ctrl+n"
 top = "home"
 bottom = "end"
 
-# style = { foreground = "", background = "", attributes = "" or [""] }
+# style = { foreground = "", background = "", attributes = "" or ["",""],  underline = "", underline_color = "" }
```
ayn2op, line 172: Unrelated change. Could you make separate these two changes into separate PR? One for fixing `mention_style` and the other for adding underline and underline_color fields to StyleWrapper.

internal/config/theme.go
xqrs: If a user have put `mention_style = { attributes = "bold" }` in their config, then the color will be blue because the style in the builtin config.toml is `mention_ = { foreground = "blue" }`. Aka the style isn't fully reset, it's appended.

The `sw.Style = tcell.StyleDefault` line will make the StyleWrapper reset on new style objects.

internal/config/config.toml
xqrs: Sure.

ayn2op (approved):

PR #765, opened 2026-03-02
fix: FreeBSD support by replacing golang.design/x/clipboard

### LINKED ISSUE #611: Cannot build on FreeBSD, since July 29
Hello,

I'm a FreeBSD Ports maintainer.
Since July 29 , discordo cannot be build on FreeBSD duo to new clipboard module (golang.design/x/clipboard) issue, that is not compatible with FreeBSD.
It is possible to revert and use the old clipboard module (github.com/atotto/clipboard)?, that is fully compatible with FreeBSD/OpenBSD/NetBSD and other UNIXes.

Thanks
~devnull

kenperkins:
## Summary
- Replace `golang.design/x/clipboard` with [`go-nativeclipboard`](https://github.com/aymanbagabas/go-nativeclipboard) which uses purego (no CGO) and has native FreeBSD support
- Define clipboard `Format` type locally instead of aliasing the library type
- Fix `internal/ui/login/form.go` which was importing `golang.design/x/clipboard` directly instead of using the internal clipboard package
- Add Wayland fallback via `wl-copy`/`wl-paste` since `go-nativeclipboard` doesn't support Wayland yet ([upstream issue](https://github.com/aymanbagabas/go-nativeclipboard/issues/7))
- Add FreeBSD cross-compilation CI job using purego's `-gcflags=-std` flag

This eliminates all platform-specific clipboard files (`clipboard_linux.go`, `clipboard_default.go`) in favor of a single `clipboard.go` with Wayland detection.

## Testing
- Verified native build on FreeBSD 15.0-RELEASE VM
- Verified build on Linux (x86_64)
- Verified clipboard functionality on macOS
- CI passes on all platforms including FreeBSD cross-compilation

Fixes #611


kenperkins:
This likely isn't ready for merge, needs to be tested across different os's. I've verified this works on MacOS, will test Linux in a few minutes.

That said, for freebsd:

https://github.com/user-attachments/assets/52015d3c-2e12-4879-bd40-9bbe594fbd21

internal/clipboard/clipboard.go
```diff
@@ -1,10 +1,94 @@
...
+	return out
+}
+
+func waylandWrite(t Format, buf []byte) {
+	cmd := exec.Command("wl-copy", "-t", waylandMIME(t))
+	cmd.Stdin = bytes.NewReader(buf)
+	if err := cmd.Run(); err != nil {
+		slog.Error("failed to write to clipboard via wl-copy", "err", err)
+	}
+}
```
ayn2op: Propagate the error returned from Run:
```go
func waylandWrite(...) error
```

internal/clipboard/clipboard.go
```diff
@@ -1,10 +1,94 @@
...
+		waylandWrite(t, buf)
+		return
+	}
+	f := formatToNative(t)
+	if _, err := f.Write(buf); err != nil {
+		slog.Error("failed to write to clipboard", "err", err)
+	}
+}
+
+func waylandRead(t Format) []byte {
```
ayn2op: Propagate the error returned from Output:
```go
func waylandRead(f Format) ([]byte, error)
```

internal/clipboard/clipboard.go
```diff
@@ -1,10 +1,94 @@
...
+func Write(t Format, buf []byte) {
+	if wayland {
+		waylandWrite(t, buf)
+		return
+	}
+	f := formatToNative(t)
+	if _, err := f.Write(buf); err != nil {
+		slog.Error("failed to write to clipboard", "err", err)
+	}
+}
```
ayn2op: Same

internal/clipboard/clipboard.go
```diff
@@ -1,10 +1,94 @@
...
+		return waylandRead(t)
+	}
+	f := formatToNative(t)
+	data, err := f.Read()
+	if err != nil {
+		slog.Error("failed to read clipboard", "err", err)
+		return nil
+	}
+	return data
+}
```
ayn2op: Same

kenperkins:
I have tested natively on FreeBSD, MacOS, and Linux (Pop 22.04) and have confirmed that discordo is functional, I don't have a Windows machine so I need some help testing there.

PR #786, opened 2026-04-01
Made Presence Status Live and also Update on Messagelist

TheMrRedSlime:
Updated the presence to change live when a person comes online or offline Since it didnt update live.
Also changes the name's color based off Presence aswell.

NOTE: Removed the Role colors from servers for this. Might want to make a toggle for it

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/alecthomas/chroma/v2 v2.23.1
 	github.com/andybalholm/brotli v1.2.1
 	github.com/ayn2op/clipboard v0.0.0-20260308203959-c5ad7df3fc97
-	github.com/ayn2op/tview v0.0.0-20260406015348-02d490edcf9c
+	github.com/ayn2op/tview v0.0.0-20260406161906-ef25e456de82
 	github.com/deckarep/gosx-notifier v0.0.0-20180201035817-e127226297fb
 	github.com/diamondburned/arikawa/v3 v3.6.1-0.20260311205148-176ad9b9440f
 	github.com/diamondburned/ningen/v3 v3.0.1-0.20260306213430-5a08d3a709b4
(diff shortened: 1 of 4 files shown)


TheMrRedSlime:
also updated it so that the dms go to the top when a person messages. live updated

TheMrRedSlime:
sorry had to fix with the new dependencies and all
had to do it manually

ayn2op:
I noted a lot of issues with this PR that need to be addressed before implementation review:
1. multiple, unrelated changes are grouped together in a single PR
2. what exactly is the PR achieves: what issue this specifically solves? could you elaborate a little in the description of the PR?
3. ideally, include a preview (screenshot) of what you have implemented.
4. I would recommend opening a GH discussion before jumping onto implementing a feature so that it can be discussed before.

TheMrRedSlime:
The main goal of this PR is to make presence status update live (currently it only shows status on initial load). I merged upstream changes first, then added live updates + status indicators next to usernames.

I'll remove the deleted message viewer from this PR so it focuses only on presence. Thanks.

Heres a summary of things i did with the fork

1. Coded the Presence statuses to update live Based on the PresenceUpdateEvent

2. Usernames on the MessagesList and DmChannelNode Update when the user goes online or offline

3. an optional config variable where The User's role color will be shown instead of the presence since that was the original behavior for the author name coloring in Guilds.

4. Changed the AppName variable for beeep to "discordo" since it defaulted to "DefaultAppName"

5. The Config File was weirdly bugged for me and did not create the config.toml file. I had to edit the code to make it create the config.toml to .config/discordo

6. When a person messaged in dms, it did not go to the top. It was a bit annoying so i coded it so that when a person messages you it will be at the top. As is normal Discord client behavior

stauersbol:
@TheMrRedSlime I think what ayyan meant is that, you have mixed a lot of things into the same PR. Wouldn't it be better to keep each of those issues in their own PRs? Because then it isolates any behaviour between the changes? Just imo not saying you have to, plus it might make the review easier (so someone is not getting confused on the changes)

PR #805, opened 2026-06-13
feat(chat): add sidebar.width_percent field for guilds tree width

DavyJonesStockings:
the width of the guildsTree sidebar was originally hardcoded so that it took up 20% of the width of the screen. this PR maintains that default value while also adding in a configuration option for users to adjust this ratio on their own.


internal/config/config.go
```diff
@@ -96,6 +96,8 @@ type (
 	}
 
 	SidebarConfig struct {
+		// Width is the ratio (%) of the guilds tree sidebar.
+		Width   int                  `toml:"width"`
```
ayn2op: Width is ambiguous because it could mean terminal columns. Since the value is explicitly a percentage, could we name this WidthPercent and describe it as such? “Ratio (%)” mixes two different representations.

internal/config/config.toml
```diff
@@ -68,6 +68,11 @@ send = true
 # Whether to receive typing status or not.
 receive = true
 
+[sidebar]
+# The relative width of the guilds tree sidebar
```
ayn2op: Could this document the unit, valid range, and invalid-value behavior? “Relative width” and “proportion” suggest a ratio or flex weight, while this setting accepts an integer percentage.
```toml
# Percentage of the available width used by the guilds tree sidebar.
# Valid values are 1-99. Invalid values fall back to 20.
width_percent = 20
```

internal/config/config.go
DavyJonesStockings: I'd be happy renaming it to WidthPercent, I agree that it's more concise and clear. Percentage is a representation of a ratio though; 20% is 20/100, which is the equivalent to the original hardcoded 1:4. If you think it's ambiguous then I'm all for rewriting the config instruction to read something along:
`// WidthPercent is the percentage (%) of the total window width that the guilds tree sidebar occupies.`

ayn2op (approved):
LGTM 🚀

PR #814, opened 2026-07-13
Added Live Presence Update

TheMrRedSlime:
Updated the Presence to change live, this is a bit of a simpler implementation and i wanted to make this a bit more neater than my previous pull request.

The Online Presence should update live for the DM Node Channel, Nothing Else.


internal/ui/chat/guilds_tree.go
```diff
@@ -74,6 +74,23 @@ func (gt *guildsTree) resetNodeIndex() {
...
+		if ch.Type == discord.DirectMessage && len(ch.DMRecipients) == 1 && ch.DMRecipients[0].ID == userID {
+			node := gt.channelNodeByID[ch.ID]
+			if node != nil {
+				gt.setNodeLineStyle(node, gt.channelNodeStyle(ch))
+				gt.chat.channelsPicker.update()
+			}
+			return
+		}
+	}
+}
```
ayn2op: Instead of using rewriting a new function to compute and update the style of the private channel (DM) node, use the `channelNodeStyle` func.

internal/ui/chat/guilds_tree.go
```diff
@@ -74,6 +74,23 @@ func (gt *guildsTree) resetNodeIndex() {
...
+	channels, err := gt.chat.state.PrivateChannels()
+	if err != nil {
+		return
+	}
+	for _, ch := range channels {
+		if ch.Type == discord.DirectMessage && len(ch.DMRecipients) == 1 && ch.DMRecipients[0].ID == userID {
+			node := gt.channelNodeByID[ch.ID]
+			if node != nil {
+				gt.setNodeLineStyle(node, gt.channelNodeStyle(ch))
+				gt.chat.channelsPicker.update()
```
ossenthusiast: Furthermore (with regards to @ayn2op's previous comment), `channelsPicker.update()` eventually calls SetItems(), which clears the active filter and resets the cursor. This means a background presence update can disrupt someone using the picker. Since picker items do not render presence state, can we remove this refresh?

TheMrRedSlime: Alright, will remove

ayn2op:
Arikawa already indexes private channels by recipient ID. Could we use `gt.chat.state.Cabinet.CreatePrivateChannel(userID)` here instead? Despite the name, the cabinet method only performs an O(1) cache lookup; unlike `State.CreatePrivateChannel`, it will not make an API request on a cache miss. This avoids sorting and scanning every private channel for each presence update.

TheMrRedSlime:
> Arikawa already indexes private channels by recipient ID. Could we use `gt.chat.state.Cabinet.CreatePrivateChannel(userID)` here instead? Despite the name, the cabinet method only performs an O(1) cache lookup; unlike `State.CreatePrivateChannel`, it will not make an API request on a cache miss. This avoids sorting and scanning every private channel for each presence update.

did you mean something to do something like this?

ayn2op:
> > Arikawa already indexes private channels by recipient ID. Could we use `gt.chat.state.Cabinet.CreatePrivateChannel(userID)` here instead? Despite the name, the cabinet method only performs an O(1) cache lookup; unlike `State.CreatePrivateChannel`, it will not make an API request on a cache miss. This avoids sorting and scanning every private channel for each presence update.
> 
> did you mean something to do something like this?

Yes

TheMrRedSlime:
> > > Arikawa already indexes private channels by recipient ID. Could we use `gt.chat.state.Cabinet.CreatePrivateChannel(userID)` here instead? Despite the name, the cabinet method only performs an O(1) cache lookup; unlike `State.CreatePrivateChannel`, it will not make an API request on a cache miss. This avoids sorting and scanning every private channel for each presence update.
> > 
> > 
> > did you mean something to do something like this?
> 
> Yes

Anything else you recommend i should do?. Or is this implementation fine enough?.

internal/ui/chat/guilds_tree.go
ossenthusiast: @TheMrRedSlime ^

TheMrRedSlime: > Instead of using rewriting a new function to compute and update the style of the private channel (DM) node, use the `channelNodeStyle` func.

@ossenthusiast @ayn2op could you clarify what you mean by this message?

ayn2op: You're right, my original comment was unclear. Your code already uses channelNodeStyle; I meant that the presence handler should only locate the cached DM channel/node and then reapply the existing style calculation, rather than introducing separate presence-style logic.

The latest implementation now does exactly that:
```go
func (gt *guildsTree) updateDMNodeStyle(userID discord.UserID) {
    channel, err := gt.chat.state.Cabinet.CreatePrivateChannel(userID)
    if err != nil {
        return
    }

    node, ok := gt.channelNodeByID[channel.ID]
    if node == nil || !ok {
        return
    }
    gt.setNodeLineStyle(node, gt.channelNodeStyle(*channel))
}
```
So this concern is resolved from my side. The lowercase method name is sufficient since it is only used within the chat package.

Also, since the method is only used within the chat package, could we make it package-private?
```go
func (gt *guildsTree) updateDMNodeStyle(userID discord.UserID) {
```
The call in onPresenceUpdate should be updated accordingly.

TheMrRedSlime: done

TheMrRedSlime: anything else i should do @ayn2op or.. is it fine now

ayn2op (approved):
Thanks! ❤️

TheMrRedSlime:
Thank you!
It is an honor to help and contribute this project. as i was using it myself a lot before creating this pull request

PR #761, opened 2026-02-27
feat(ui/chat): add user context menu

kenperkins:
## Summary
- Add a hotkey-driven user context menu triggered via `Ctrl+P` from the messages list or guilds tree (DM nodes only)
- Menu items: **Profile** (P), **Message** (M), and **Note** (N) — executed by pressing the highlighted letter
- Message option only shown for existing friends or DM channels to avoid Discord captcha issues
- Profile shows an overlay with user info (display name, username, ID, created date, roles, note)
- Note opens a form to view/edit the user's note

CleanShot 2026-02-27 at 08 43 19

https://github.com/user-attachments/assets/c5747b0d-8e8e-42c7-9eb1-d5683cd76011

## Test plan
- [x] Build: `go build ./...`
- [x] Select a message in messages list, press `Ctrl+P` — menu appears anchored near the message
- [x] Select a DM channel in guilds tree, press `Ctrl+P` — menu appears for that DM user
- [x] Press `P` to view profile overlay, `Esc` to close
- [x] Press `N` to edit note, save/cancel works
- [x] Press `M` to open DM channel (only visible for friends/existing DMs)
- [x] Press `Esc` to dismiss menu
- [x] Verify help bar updates when menu is visible
- [x] Verify `Ctrl+P` on non-DM guilds tree nodes is a no-op

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #244, opened 2022-10-15
see user status and info in user menu


sewnie:
to tell the person in direct messages if they are online, offline, dnd, idle, join date, join date in server, etc etc.

kenperkins:
This might be worth adding onto #761 if that lands.

PR #818, opened 2026-07-21
Status Indicator in Message List

TheMrRedSlime:
Indicate the user's status on the message list. Seeing when they are offline, online, idle, dnd, etc.

The Indicator can be customized to be anything on the config file.
It also Syncs with the presence update Event

### DIFF
--- a/internal/ui/chat/messages_list.go
+++ b/internal/ui/chat/messages_list.go
@@ -368,47 +368,42 @@ func (ml *messagesList) drawTimestamps(builder *tview.LineBuilder, ts discord.Ti
 	builder.Write(ml.formatTimestamp(ts)+" ", dimStyle)
 }
 
-func (ml *messagesList) getStatusIndicator(message discord.Message) string {
+func (ml *messagesList) statusIndicator(message discord.Message) string {
 	if message.WebhookID.IsValid() {
 		return ""
 	}
 
 	guildID := message.GuildID
-
 	if !guildID.IsValid() {
-		guildID = discord.NullGuildID // Fallback to null guild ID for DMs and group DMs
+		guildID = discord.NullGuildID
 	}
 
 	presence, err := ml.chat.state.Cabinet.Presence(guildID, message.Author.ID)
 	if err != nil {
-		return ""
+		return ml.cfg.Theme.MessagesList.StatusIndicators[string(discord.OfflineStatus)]
 	}
 
-	var statusKey string
-	switch presence.Status {
-	case discord.DoNotDisturbStatus:
-		statusKey = "dnd"
-	case discord.IdleStatus:
-		statusKey = "idle"
-	case discord.OnlineStatus:
-		statusKey = "online"
-	default:
-		statusKey = "offline"
+	status := presence.Status
+	if status == discord.UnknownStatus || status == discord.InvisibleStatus {
+		status = discord.OfflineStatus
 	}
-
-	if indicator, ok := ml.cfg.Theme.MessagesList.StatusIndicators[statusKey]; ok {
-		return indicator
-	}
-	return ""
+	return ml.cfg.Theme.MessagesList.StatusIndicators[string(status)]
 }
 
 func (ml *messagesList) updateUserMessages(userID discord.UserID) {
-	for _, msg := range ml.messages {
-		if (msg.Author.ID == userID) || (msg.ReferencedMessage != nil && msg.ReferencedMessage.Author.ID == userID) {
-			if tv, ok := ml.itemByID[msg.ID]; ok {
-				newLines := ml.renderMessage(msg, ml.cfg.Theme.MessagesList.MessageStyle.Style)
-				tv.SetLines(newLines)
-			}
+	style := ml.cfg.Theme.MessagesList.MessageStyle.Style
+
+	for _, message := range ml.messages {
+		isAuthor := message.Author.ID == userID
+		isReplyAuthor := message.ReferencedMessage != nil &&
+			message.ReferencedMessage.Author.ID == userID
+
+		if !isAuthor && !isReplyAuthor {
+			continue
+		}
+
+		if view := ml.itemByID[message.ID]; view != nil {
+			view.SetLines(ml.renderMessage(message, style))
 		}
 	}
 }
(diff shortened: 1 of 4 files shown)


internal/ui/chat/messages_list.go
```diff
@@ -368,6 +368,51 @@ func (ml *messagesList) drawTimestamps(builder *tview.LineBuilder, ts discord.Ti
...
+	if !guildID.IsValid() {
+		guildID = discord.NullGuildID // Fallback to null guild ID for DMs and group DMs
+	}
+
+	presence, err := ml.chat.state.Cabinet.Presence(guildID, message.Author.ID)
+	if err != nil {
+		return ""
+	}
+
+	var statusKey string
```
ayn2op: `discord.Status` is already a string-backed enum whose values match these configuration keys, so this switch can be simplified to statusStr := string(presence.Status). Only `UnknownStatus` and `InvisibleStatus` need explicit normalization to `OfflineStatus`. Also, consider treating a missing presence record as offline, since offline users frequently have no cached presence and currently return no indicator.

```go
status := presence.Status
if status == discord.UnknownStatus || status == discord.InvisibleStatus {
    status = discord.OfflineStatus
}
return ml.cfg.Theme.MessagesList.StatusIndicators[string(status)]
```

internal/ui/chat/messages_list.go
```diff
@@ -368,6 +368,51 @@ func (ml *messagesList) drawTimestamps(builder *tview.LineBuilder, ts discord.Ti
...
+}
+
+func (ml *messagesList) updateUserMessages(userID discord.UserID) {
+	for _, msg := range ml.messages {
+		if (msg.Author.ID == userID) || (msg.ReferencedMessage != nil && msg.ReferencedMessage.Author.ID == userID) {
+			if tv, ok := ml.itemByID[msg.ID]; ok {
+				newLines := ml.renderMessage(msg, ml.cfg.Theme.MessagesList.MessageStyle.Style)
+				tv.SetLines(newLines)
+			}
+		}
```
ayn2op: Remove redundant parentheses and temporary variable. Go conditionals don't require parentheses, and `newLines` is only used once.

internal/ui/chat/messages_list.go
```diff
@@ -368,6 +368,51 @@ func (ml *messagesList) drawTimestamps(builder *tview.LineBuilder, ts discord.Ti
 	builder.Write(ml.formatTimestamp(ts)+" ", dimStyle)
 }
 
+func (ml *messagesList) getStatusIndicator(message discord.Message) string {
```
ayn2op: Follow Go naming conventions; avoid the `get` prefix: `func (ml *messagesList) statusIndicator(message discord.Message) string`

internal/ui/chat/messages_list.go
```diff
@@ -388,6 +433,13 @@ func (ml *messagesList) drawAuthor(builder *tview.LineBuilder, message discord.M
 
 	style := baseStyle.Foreground(foreground).Bold(true)
 	builder.Write(name+" ", style)
+
+	if indicator := ml.getStatusIndicator(message); indicator != "" {
+		whiteStyle := baseStyle.Foreground(color.White)
```
ayn2op: `whiteStyle` describes a specific color rather than its purpose:
```go
indicatorStyle := baseStyle.Foreground(color.White)
```

ayn2op: Or `delimiterStyle`, since it is only used for the parentheses.

internal/ui/chat/messages_list.go
```diff
@@ -388,6 +433,13 @@ func (ml *messagesList) drawAuthor(builder *tview.LineBuilder, message discord.M
...
+		builder.Write("(", whiteStyle)
+		builder.Write(indicator, style)
+		builder.Write(") ", whiteStyle)
```
ayn2op: The three writes are unnecessary if the indicator does not intentionally need the author's role color: `builder.Write("("+indicator+") ", indicatorStyle)`

If the differently styled indicator is intentional, keep the separate writes.

internal/ui/chat/messages_list.go
```diff
@@ -368,6 +368,51 @@ func (ml *messagesList) drawTimestamps(builder *tview.LineBuilder, ts discord.Ti
...
+	case discord.OnlineStatus:
+		statusKey = "online"
+	default:
+		statusKey = "offline"
+	}
+
+	if indicator, ok := ml.cfg.Theme.MessagesList.StatusIndicators[statusKey]; ok {
+		return indicator
+	}
+	return ""
```
ayn2op: Consider returning the configured offline indicator when presence is absent:
```go
  presence, err := ml.chat.state.Cabinet.Presence(guildID, message.Author.ID)
  if err != nil {
  	return ml.cfg.Theme.MessagesList.StatusIndicators[string(discord.OfflineStatus)]
  }
```
This shortens the later logic and ensures offline users receive an indicator.

internal/ui/chat/messages_list.go
```diff
@@ -368,6 +368,51 @@ func (ml *messagesList) drawTimestamps(builder *tview.LineBuilder, ts discord.Ti
...
+}
+
+func (ml *messagesList) updateUserMessages(userID discord.UserID) {
```
ayn2op, line 393:
```go
func (ml *messagesList) updateUserMessages(userID discord.UserID) {
  	style := ml.cfg.Theme.MessagesList.MessageStyle.Style

  	for _, message := range ml.messages {
  		isAuthor := message.Author.ID == userID
  		isReplyAuthor := message.ReferencedMessage != nil &&
  			message.ReferencedMessage.Author.ID == userID

  		if !isAuthor && !isReplyAuthor {
  			continue
  		}

  		if view := ml.itemByID[message.ID]; view != nil {
  			view.SetLines(ml.renderMessage(message, style))
  		}
  	}
 }
```

internal/ui/chat/messages_list.go
TheMrRedSlime: It was meant to be intentional, but now that i think about it i should keep everything in one style could just simplify the indicator

internal/ui/chat/messages_list.go
TheMrRedSlime: done

internal/ui/chat/messages_list.go
TheMrRedSlime: also done

internal/ui/chat/messages_list.go
TheMrRedSlime: done

TheMrRedSlime:
Did some changes n fixes from what you said. uh anything else?

ayn2op:
image

Doesn't work. Everyone's presence is shown as offline. Furthermore, I believe (<status>) is extremely distracting and should be moved somewhere else. I have thought about adding a floating modal for user's entire profile.

TheMrRedSlime:
> image
> 
> Doesn't work. Everyone's presence is shown as offline. Furthermore, I believe (<status>) is extremely distracting and should be moved somewhere else. I have thought about adding a floating modal for user's entire profile.

Should i just remove the brackets if you feel it is distracting?
Lemme see what's going on.

ayn2op:
Related: https://github.com/ayn2op/discordo/issues/244

ayn2op:
Related: https://github.com/ayn2op/discordo/pull/761

PR #831, opened 2026-08-27
fix(ui/chat): show live threads and open channels that have them

### LINKED ISSUE #309: Threads not shown
I've joined some continuous threads, but these threads are not shown in discordo.

Discord Official FAQ about Threads:
https://support.discord.com/hc/ja/articles/4403205878423-Threads-FAQ

Cephalode:
Fixes #309

- `onSelected` toggles by expandability+loaded instead of raw child count, so a channel holding synced thread children still opens on Enter
- handle THREAD_CREATE / THREAD_UPDATE / THREAD_DELETE / THREAD_LIST_SYNC: insert, restyle, archive-remove, and de-duplicating re-sync of thread nodes under their parent channel
- skip the ViewChannel permission gate for threads (`CalcOverrides` reads the thread's own empty overwrites, not the parent's) and drop archived threads
- channels picker: hide thread entries

### DIFF
--- a/internal/ui/chat/state.go
+++ b/internal/ui/chat/state.go
@@ -244,6 +244,69 @@ func (m *Model) onTypingStart(event *gateway.TypingStartEvent) {
 	m.addTyper(event.UserID)
 }
 
+func (m *Model) onThreadCreate(thread discord.Channel) {
+	if !isThread(thread.Type) {
+		return
+	}
+	if thread.ParentID.IsValid() {
+		if parentNode := m.guildsTree.findNodeByReference(thread.ParentID); parentNode != nil {
+			m.guildsTree.createChannelNode(parentNode, thread)
+		}
+	}
+}
+
+func (m *Model) onThreadUpdate(thread discord.Channel) {
+	if !isThread(thread.Type) {
+		return
+	}
+	if node := m.guildsTree.findNodeByReference(thread.ID); node != nil {
+		if thread.ThreadMetadata != nil && thread.ThreadMetadata.Archived {
+			if parentNode := m.guildsTree.findNodeByReference(thread.ParentID); parentNode != nil {
+				parentNode.RemoveChild(node)
+			}
+			delete(m.guildsTree.channelNodeByID, thread.ID)
+		} else {
+			m.guildsTree.setNodeLineStyle(node, m.guildsTree.channelNodeStyle(thread))
+		}
+	}
+}
+
+func (m *Model) onThreadDelete(thread gateway.ThreadDeleteEvent) {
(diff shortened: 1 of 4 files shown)


ayn2op:
Thanks for working on this. [`THREAD_LIST_SYNC`](https://docs.discord.food/gateway/gateway-events#thread-list-sync) has a non-obvious edge case.

The event is authoritative for the synchronized scope. channel_ids can contain a parent with no active threads specifically so clients know to clear its cached threads. When `channel_ids` is omitted, the scope is the entire guild.

A straightforward approach would be:

1. Determine the affected parents, or use the whole guild when `channel_ids` is nil.
2. Remove existing thread nodes in that scope.
3. Add the active threads from `event.Threads`.

[discord.py-self](https://github.com/dolfies/discord.py-self) removes unseen cached threads and then adds or updates the returned threads.

internal/ui/chat/state.go
```diff
@@ -244,6 +244,69 @@ func (m *Model) onTypingStart(event *gateway.TypingStartEvent) {
...
+				parentNode.RemoveChild(node)
+			}
+			delete(m.guildsTree.channelNodeByID, thread.ID)
+		} else {
+			m.guildsTree.setNodeLineStyle(node, m.guildsTree.channelNodeStyle(thread))
+		}
+	}
+}
+
+func (m *Model) onThreadDelete(thread gateway.ThreadDeleteEvent) {
```
ayn2op, line 274: Nice start on the event handling. I think `CREATE` and `UPDATE` can share a small upsert path, which also handles a few gateway edge cases.

`THREAD_CREATE` can arrive for a thread already in memory. `THREAD_UPDATE` can rename, archive, or unarchive a thread. Currently `CREATE` may add a duplicate node, while `UPDATE` only changes its style and cannot restore an unarchived thread or update its displayed name.

One simple structure could be:

- `upsertThread(thread)`: Remove any existing node, then add the active thread under its parent.
- `removeThread(id, parentID)`: Remove the node and index entry.
- `CREATE` and active `UPDATE` call `upsertThread`.
- Archived `UPDATE` and `DELETE` call `removeThread`.
- `LIST_SYNC` clears its scope, then calls `upsertThread` for every returned thread.

[Read more about thread synchronization](https://docs.discord.food/topics/threads#details-about-thread-access-and-syncing)

ayn2op:
[pr-831-thread-sync.patch](https://github.com/user-attachments/files/31498183/pr-831-thread-sync.patch)

internal/ui/chat/state.go
Cephalode: My bad, I overlooked the possible node duplication. I will do a full edge case combination test before posting an posting an update.

